PR #2, opened 2022-10-25
global: update deps

jrcastro2:


Dockerfile
```diff
@@ -8,7 +8,7 @@
 # Note: It is important to keep the commands in this file in sync with your
 # bootstrap script located in ./scripts/bootstrap.
 
-FROM inveniosoftware/centos8-python:3.8
+FROM inveniosoftware/centos7-python:3.9
```
zzacharo, line 11: wouldn't that point to almalinux?

zzacharo (approved):

Dockerfile
jrcastro2: Unless I am missing something it would use centos7 and python 3.9, see [here](https://github.com/inveniosoftware/docker-invenio). Actually I don't think we have already published a version with almalinux, see [dockerhub here](https://hub.docker.com/search?q=inveniosoftware%2F)

PR #51, opened 2023-01-09
Records list: refactor override to use only jsx

Pineirin:
Hide section on empty results.
closes: inveniosoftware/invenio-app-rdm#1996

Screenshot from 2023-01-09 14-23-19

### DIFF
--- a/assets/js/components/frontpage/overrides/RecordsList.js
+++ b/assets/js/components/frontpage/overrides/RecordsList.js
@@ -7,9 +7,9 @@
 import _get from "lodash/get";
 import _truncate from "lodash/truncate";
 import _upperCase from "lodash/upperCase";
-import React, { Component } from "react";
+import React from "react";
 import PropTypes from "prop-types";
-import { withCancel, http } from "react-invenio-forms";
+import isEmpty from "lodash/isEmpty";
 import {
   Loader,
   Container,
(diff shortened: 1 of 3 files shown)


assets/js/components/frontpage/overrides/RecordsList.js
```diff
@@ -18,60 +18,10 @@ import {
...
-    const { title } = this.props;
-
-    const listItems = data.hits?.map((record) => {
-      return <RecordsResultsListItem result={record} key={record.id} />;
-    });
-
-    return (
+export const CDSRecordsList = ({ isLoading, error, title, listItems }) => {
+  return (
+    !isEmpty(listItems) && (
```
kpsherva, line 24: wouldn't it better if we implement this globally so that other instances (f.e. Zenodo) can also benefit from this fix ?

Pineirin: This was implemented on invenio-app-rdm, so its global. I left the if condition inside the overridable jsx in case in some instance we want to display a message instead of completely removing the section.

jrcastro2 (approved):

assets/js/components/frontpage/overrides/RecordsList.js
kpsherva: could you show me there exactly this if statement is implemented in app_rdm ?

Pineirin: https://github.com/inveniosoftware/invenio-app-rdm/pull/2007/files#diff-bc5bd180fc17fe701ef603df6390ca4466032be536713855b1fb469b13c2e510R72

Issue #16, opened 2022-11-23
[DEPENDS ON #invenio-app-rdm/2186] CERN e-groups integration


ntarocco:
- [x] make sure that the logout redirects to CERN logout (see [here](https://auth.docs.cern.ch/user-documentation/oidc/config/#logout) related docs)
- [x] implement groups handler for keycloak to fetch my groups

* DEPENDS ON https://github.com/inveniosoftware/invenio-app-rdm/issues/2186

See https://codimd.web.cern.ch/GyeRhRCORumc-8ak0t26Jg

## CERN login

For the time being, we can probably disable the fetching of the user info endpoint (we don't need any extra info).
`OAUTHCLIENT_CERN_USER_INFO_FROM_ENDPOINT = False`

## Fetch groups

Use the endpoint: `/api/v1.0/IdentityMembership/{cern_upn}/precomputed`, see example [here](https://github.com/indico/flask-multipass-cern/blob/master/flask_multipass_cern.py#L145).

PR #66, opened 2023-05-23
oidc: adds group handler

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/16

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -194,14 +194,26 @@ SECURITY_SEND_REGISTER_EMAIL = False
 # -------------------
 # See https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/config.py
 from invenio_oauthclient.contrib.keycloak import KeycloakSettingsHelper
-from cds_rdm.oidc import cern_info_serializer, confirm_registration_form
+from cds_rdm.oidc import (
+    cern_info_serializer,
+    confirm_registration_form,
+    cern_group_serializer,
+    cern_group_handler,
+)
+from urllib.parse import quote
+
+_base_url = os.environ.get("CERN_KEYCLOAK_BASE_URL", "https://keycloak-qa.cern.ch/")
 
 _keycloak_helper = KeycloakSettingsHelper(
     title="CERN",
     description="CERN SSO authentication",
-    base_url=os.environ.get("CERN_KEYCLOAK_BASE_URL", "https://keycloak-qa.cern.ch/"),
+    base_url=_base_url,
     realm="cern",
     app_key="CERN_APP_CREDENTIALS",
+    logout_url="{}auth/realms/cern/protocol/openid-connect/logout?redirect_uri={}".format(
+        _base_url,
+        quote(os.environ.get("INVENIO_SITE_UI_URL", SITE_UI_URL))
+    ),
 )
 OAUTHCLIENT_CERN_REALM_URL = _keycloak_helper.realm_url
 OAUTHCLIENT_CERN_USER_INFO_URL = _keycloak_helper.user_info_url
@@ -210,9 +222,19 @@ OAUTHCLIENT_CERN_VERIFY_AUD = False
 OAUTHCLIENT_CERN_USER_INFO_FROM_ENDPOINT = False
 
 handlers = _keycloak_helper.get_handlers()
-handlers["signup_handler"]["info_serializer"] = cern_info_serializer
-rest_handlers = _keycloak_helper.get_handlers()
-rest_handlers["signup_handler"]["info_serializer"] = cern_info_serializer
+handlers["signup_handler"] = {
+    **handlers["signup_handler"],
+    "info_serializer": cern_info_serializer,
+    "groups_serializer": cern_group_serializer,
+    "groups": cern_group_handler,
+}
+rest_handlers = _keycloak_helper.get_rest_handlers()
+rest_handlers["signup_handler"] = {
+    **rest_handlers["signup_handler"],
+    "info_serializer": cern_info_serializer,
+    "groups_serializer": cern_group_serializer,
+    "groups": cern_group_handler,
+}
 
 OAUTHCLIENT_SIGNUP_FORM = confirm_registration_form
 
--- a/site/cds_rdm/oidc.py
+++ b/site/cds_rdm/oidc.py
@@ -3,21 +3,43 @@
 from invenio_userprofiles.forms import confirm_register_form_preferences_factory
 from flask import current_app
 from werkzeug.local import LocalProxy
+from invenio_oauthclient import current_oauthclient
+from invenio_oauthclient.contrib.keycloak.handlers import get_user_info
 
 _security = LocalProxy(lambda: current_app.extensions["security"])
 
 
 def confirm_registration_form(*args, **kwargs):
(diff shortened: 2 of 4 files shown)


invenio.cfg
```diff
@@ -211,15 +223,23 @@ OAUTHCLIENT_CERN_USER_INFO_FROM_ENDPOINT = False
...
 OAUTHCLIENT_SIGNUP_FORM = confirm_registration_form
 
+_remote_app_name = "cern"
+
 OAUTHCLIENT_REMOTE_APPS = {
-    "cern": _keycloak_helper.remote_app,
+    _remote_app_name: _keycloak_helper.remote_app,
 }
 
+SECURITY_POST_LOGOUT_VIEW = "/oauth/{}/logout".format(_remote_app_name)
```
jrcastro2, line 241: This works only for one remote, any app with more remotes should be able register it's own logout url, to be checked if it makes sense to override ``SECURITY_LOGOUT_URL``

invenio.cfg
```diff
@@ -194,14 +194,26 @@ SECURITY_SEND_REGISTER_EMAIL = False
...
 _keycloak_helper = KeycloakSettingsHelper(
     title="CERN",
     description="CERN SSO authentication",
-    base_url=os.environ.get("CERN_KEYCLOAK_BASE_URL", "https://keycloak-qa.cern.ch/"),
+    base_url=_base_url,
     realm="cern",
     app_key="CERN_APP_CREDENTIALS",
+    logout_url="{}auth/realms/cern/protocol/openid-connect/logout?redirect_uri={}".format(
+        _base_url,
+        quote(SITE_UI_URL)
```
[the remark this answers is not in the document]
kpsherva: +1 on this, I think if you try to replace the value by env variable it wouldn't work because it would take the one from this file instead.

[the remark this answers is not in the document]
slint: One alternative is to use a "lazy"/LocalProxy to get the value from `current_app.config` like so:

```python
logout_url=LocalProxy(
    lambda: "{}auth/realms/cern/protocol/openid-connect/logout?redirect_uri={}".format(
        _base_url,
        quote(current_app.config["SITE_UI_URL"])
    )
```

Not sure if this will work though, depends on when `logout_url` is accessed (hopefully not outside application context).

slint (commented):
One minor comment on the config loading as well which might save relying on `os.environ` and use the Invenio config.

invenio.cfg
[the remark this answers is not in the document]
ppanero: From review along with @anikachurilova +1 +1 to needs lazy loading

invenio.cfg
```diff
@@ -211,8 +223,12 @@ OAUTHCLIENT_CERN_USER_INFO_FROM_ENDPOINT = False
 
 handlers = _keycloak_helper.get_handlers()
 handlers["signup_handler"]["info_serializer"] = cern_info_serializer
-rest_handlers = _keycloak_helper.get_handlers()
-rest_handlers["signup_handler"]["info_serializer"] = cern_info_serializer
+handlers["signup_handler"]["groups_serializer"] = cern_group_serializer
```
ppanero, line 226: nit for readability:

```python
handlers["signup_handler"] = {
    info_serializer: ...
    groups_serializer: ...
    ...
}
```

carefully with the fact that dictionaries are references (not copies), but you can even reuse the dict above for both rest and ui handlers. But copy might actually be ugly, so the above for readability.

site/cds_rdm/oidc.py
```diff
@@ -3,21 +3,43 @@
...
+    for group_name in groups:
+        serialized_groups.append({"id": group_name, "name": group_name})
+
+    return serialized_groups
+
 
+def cern_group_handler(remote, resp, user):
+    """Retrieves groups from remote account."""
+    token_user_info, user_info = get_user_info(remote, resp)
+    groups = token_user_info.get("groups", [])
```
ppanero, line 36: as we do for other values in the [keycloak serializer](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/contrib/keycloak/handlers.py#LL65C2-L65C63), would it make sense to check the `user_info` if the token does not contain groups? I know CERN's IdP gives them in the token, but just to be safe on changes.

site/cds_rdm/oidc.py
```diff
@@ -3,21 +3,43 @@
...
+
     return _Form(*args, **kwargs)
 
 
-def cern_info_serializer(remote, resp, token_user_info, user_info):
+def cern_group_serializer(remote, groups, **kwargs):
+    """Serialize the groups response object."""
+    serialized_groups = []
+    # E-groups do have unique names and this name cannot be updated, therefore the name can act as an ID for invenio
+    for group_name in groups:
```
ppanero, line 27: why does this have a different format than https://github.com/CERNDocumentServer/cds-rdm/pull/74/commits/932157820d40572ccb53caeffacb58b0d32f1be6#diff-48d7dfeba7dedee600381f6c5491916254d518942b2c2708f42135df63e3a690R59

jrcastro2: Good question!

Because the information we get is different, when syncing user e-groups on log in, we only get a list of identifiers (`["zenodo-admins", cds-admins", ...]`), so we set the id with that value and the name with the same value (which in CERN e-groups is always the same). 

The referenced PR you have pasted (group syncing cronjob), does fetch all the groups with all the information, therefore, we will also get the description (and the name in a separate attribue although it's the same as the id), so that we can complete the information of the groups. 

This was agreed with Nico and the auth team.

site/cds_rdm/oidc.py
jrcastro2: From my understanding in that serializer we do check for the information reuqired to create the user profile. Not sure what are the benefits of adding a check for the groups there?

invenio.cfg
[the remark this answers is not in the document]
jrcastro2: Thanks for pointing this out! 
LocalProxy doesn't seem to work fine, I've used `os.environ`, unless someone has a better suggestion.
Screenshot from 2023-06-19 18-58-54

site/cds_rdm/oidc.py
ppanero: Is not about adding a check, is about safety for future changes. Having the groups in the OAuth token response is not standard (IIRC it was requested for the CERN Auth team to add them). Usually, they come in the user info request. Which means by "standard" they would come in `user_info`. Since this is `CDS-RDM` it okay to leave it as is (sorry, should've maked the comment as minor). My point was about doing something like:

```python
groups = token_user_info.get("groups", [])
if not groups:
    groups = user_info.get("groups", [])
```

In case the implementation changes back to standard in the future. But is ok to leave it as is.

invenio.cfg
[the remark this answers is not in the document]
slint: Oof, this was indeed really "lazy" all the way until before adding the redirect header 🤦 

Ok, last suggestion: could we make that the `logout_url` config is a string OR function/callable? In that case it's "lazy" up to the point of actually performing the redirect in the `/oauth/logout` endpoint, where you could call the function.

This is not blocking for merging at all, so we can open also just open an issue.

site/cds_rdm/oidc.py
kpsherva: I created a ticket to improve

PR #73, opened 2023-06-08
assets: add changes necessary for less upgrade

kpsherva:
Screenshot 2023-06-12 at 10-08-32 CDS


slint (commented):
LGTUs, but we're not that much of LESS experts to say for sure 😅

PR #75, opened 2023-06-16
Add user sync

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/67
* NOTE FOR REVIEWERS: Only review last PR


site/cds_rdm/ldap/user_importer.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self):
+        """Constructor."""
+        self.client_id = current_app.config["CERN_APP_CREDENTIALS"]["consumer_key"]
+
+    def create_invenio_user(self, ldap_user):
+        """Commit new user in db."""
+        email = ldap_user["user_email"]
+        username = ldap_user["keycloak_id"]
+        user = User(email=email, username=username, active=True)
```
jrcastro2, line 40: Here we set the username to the keycloak_id/ldpa cn, just want to double check that it's okay to expose this as the username

site/cds_rdm/ldap/user_importer.py
```diff
@@ -0,0 +1,84 @@
...
+    def create_invenio_user_profile(self, user, ldap_user):
+        """Return new user profile."""
+        user_profile = UserProfile(user=user)
+        user_profile.full_name = ldap_user["user_profile_full_name"]
+        return user_profile
+
+    def create_invenio_remote_account(self, user_id, ldap_user):
+        """Return new user entry."""
+        keycloak_id = ldap_user["keycloak_id"]
+        department = ldap_user["remote_account_department"]
```
jrcastro2, line 63: TBC if we should keep the deparment information or not

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+            # get first value of each field from LDAP
+            # decode bytes to UTF-8
+            decoded_data[key] = value[0].decode("utf8")
+        serialized_data = dict(
+            user_email=decoded_data["mail"].lower(),
+            user_profile_full_name=decoded_data["displayName"],
+            user_identity_id=decoded_data["uidNumber"],
+            cern_account_type=decoded_data["cernAccountType"],
+            remote_account_department=decoded_data["department"],
+            keycloak_id=decoded_data.get("name", decoded_data["cn"]),
```
jrcastro2, line 33: I am not 100% sure about this, maybe we should rely only on one, the `cn` or `name`, prior agreement wit the Auth team to make sure which one will map to the keycloak id all the time

invenio.cfg
```diff
@@ -182,6 +184,7 @@ SECURITY_RECOVERABLE = False  # local login: allow users to reset the password
 SECURITY_CHANGEABLE = False  # local login: allow users to change psw
 SECURITY_CONFIRMABLE = False  # local login: users can confirm e-mail address
 SECURITY_LOGIN_WITHOUT_CONFIRMATION = True # require users to confirm email before being able to login
+ACCOUNTS_DEFAULT_USER_VISIBILITY = "public" # enables users to be searchable for invites
```
jrcastro2, line 187: This will enable users to allow to search for them by name, however there is another variable to allow to search by email (and in consequence exposes the email) do we want to allow this as well?

In another note, at the moment in CDS-RDM we are not confirming the users when syncing or logging in, there is actually no way to confirm the account. Without confirming the acount the users are not searchable, if I am not mistaken it's part of the logic to make it searchable. Should we automatically confirm the users internally (when logging in and syncing)? Should we enable the normal confirmation (if we do this, synced users won't be searchable as they have to first confirm the accounts)? 

Alternatively we could change the logic to not require user confirmation to make it searchable. 

ping @kpsherva

site/cds_rdm/ldap/user_importer.py
```diff
@@ -0,0 +1,84 @@
...
+        user = User(email=email, username=username, active=True)
+        db.session.add(user)
+        db.session.commit()
+        return user
+
+    def create_invenio_user_identity(self, user_id, ldap_user):
+        """Return new user identity entry."""
+        uid_number = ldap_user["user_identity_id"]
+        return UserIdentity(
+            id=uid_number,
```
jrcastro2, line 49: Minor: We can use the employeeId and no need to fetch the uidNumber

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+        # check if email is empty string.
+        # It happens when the account is not fully created on LDAP
+        if not email:
+            log_func_missing_email = partial(
+                log_func, extra=dict(employee_id=employee_id)
+            )
+            raise InvalidLdapUser(
+                f"LDAP user with employeeID {employee_id}" f" has no email address.",
+                log_func=log_func_missing_email,
+            )
```
jrcastro2, line 60: Minor: Ths 2 functions can be combined into 1

jrcastro2 (commented):
Peer reviewed with @psaiz

site/cds_rdm/tasks.py
```diff
@@ -0,0 +1,17 @@
...
+from cds_rdm.ldap.api import update_users
+from flask import current_app
+from invenio_db import db
+
+
+@shared_task
+def synch_users():
+    """Run the task to update users from LDAP."""
+    try:
+        update_users()
```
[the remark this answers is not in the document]
kpsherva: as far as I understood from Nico no, as long as we have the emails

site/cds_rdm/ldap/user_importer.py
kpsherva: it is already exposed in the webservices internally if I am not mistaken, also via e-groups

site/cds_rdm/ldap/user_importer.py
```diff
@@ -0,0 +1,84 @@
...
+            method=current_app.config.get("OAUTH_REMOTE_APP_NAME"),
+            id_user=user_id,
+        )
+
+    def create_invenio_user_profile(self, user, ldap_user):
+        """Return new user profile."""
+        user_profile = UserProfile(user=user)
+        user_profile.full_name = ldap_user["user_profile_full_name"]
```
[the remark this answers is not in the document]
kpsherva: I would leave it as is in ILS for now. We might need the affiliations though later, when assigning authors, therefore I would create a ticket for this comment

site/cds_rdm/ldap/user_importer.py
kpsherva: we should since we have it

invenio.cfg
kpsherva: to be verified that public visibility means only visible for logged in users (not 100% public)

invenio.cfg
```diff
@@ -262,3 +267,18 @@ OAISERVER_ID_PREFIX = "cds-rdm.com"
 
 SEARCH_INDEX_PREFIX = "cds-rdm-"
 COMMUNITIES_ALLOW_RESTRICTED = False
+
+
+###############################################################################
+# CDS-RDM configuration
+###############################################################################
+#: LDAP configuration
+CDS_RDM_LDAP_URL = "ldap://xldap.cern.ch"
```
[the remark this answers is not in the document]
kpsherva: any env variable is possible to be set if you add `INVENIO_` as a prefix if I remember correctly

invenio.cfg
jrcastro2: If not logged in we get a 403 when fetching the users API (api/users)

site/setup.cfg
```diff
@@ -3,6 +3,14 @@
 name = cds-rdm
 version = attr: cds_rdm.__version__
 
+[options]
+include_package_data = True
+packages = find:
+python_requires = >=3.6
```
ppanero, line 9: I think 3.7 is reaching EOL and security fixes, maybe upgrade to 3.8? or even 3.9 since the RDM images with AlmaLinux are 3.9 only?

site/cds_rdm/tasks.py
```diff
@@ -111,3 +113,13 @@ def update_groups(self, offset, limit, groups_headers):
             }
         )
     create_or_update_groups(serialized_groups)
+
+
+@shared_task
```
ppanero, line 118: do you need retries or similar? or is that done at celery configuration level.

site/cds_rdm/oidc.py
```diff
@@ -57,7 +79,7 @@ def cern_info_serializer(remote, resp, token_user_info, user_info):
                 "username": username,
             },
             "prefs": {
-                "visibility": "restricted",
+                "visibility": "public",
```
ppanero, line 82: worth using the config variable? `ACCOUNTS_DEFAULT_USER_VISIBILITY`

site/cds_rdm/ldap/errors.py
```diff
@@ -0,0 +1,21 @@
...
+import sys
+
+
+class InvalidLdapUser(Exception):
+    """Invalid user exception."""
+
+    def __init__(self, *args, log_func=None):
+        """Constructor."""
+        super().__init__(*args)
+        print(args[0], file=sys.stderr)
```
ppanero, line 19: 😏 
```suggestion
```

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+from functools import partial
+
+from cds_rdm.ldap.errors import InvalidLdapUser
+from invenio_accounts.models import User
+from invenio_oauthclient.models import UserIdentity
+from invenio_userprofiles import UserProfile
+
+
+def serialize_ldap_user(ldap_user_data, log_func=None):
+    """Create ldap user."""
```
ppanero, line 19:
```suggestion
    """Serialize ldap user."""
```

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def serialize(ldap_user_data):
```
ppanero, line 21: several comments on the serialization:
- The two validation functions seem a bit redundant, since they check for the same. While the second one checks for an empty string, I think checking for `""` should also be part of the required because an empty email is a wrong response form the IdP/LDAP. Note that even the error message are identical.
- Then in lines 26-33 all keys on the dictionary are accessed with `[]` which means they are required (or it will raise a `KeyError`). Should they be accessed with `.get()`?

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+        return
+
+
+def user_exists(ldap_user):
+    """Check if user exists in the db."""
+    if not ldap_user:
+        return False
+
+    email = ldap_user["user_email"]
+    user_email_exists = User.query.filter_by(email=email).count() > 0
```
ppanero, line 78: I _think_ a `count` query is more expensive than a `one_or_none` since it will do a full pass on the db table. `email` has a unique constraint in invenio accounts so there could not be more than one.

To be checked, does sqlalchemy ORM have an `exists()`, it might be more optimized for this use case.

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if user_email_exists:
+        return True
+
+    user_identity_exists = (
+        UserIdentity.query.filter_by(id_user=ldap_user["user_identity_id"]).count() > 0
```
ppanero, line 84: same as above.

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+    if user_identity_exists:
+        return True
+
```
ppanero, line 88: should it explicitly return `False` at the end?

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+class InvenioUser:
+    """Invenio user serializer class."""
+
+    def __init__(self, remote_account):
+        """Constructor."""
+        self.user_id = remote_account.user_id
+        self.remote_account = remote_account
+        self.user_profile = UserProfile.get_by_userid(self.user_id)
+        self.user_identity = UserIdentity.query.filter_by(id_user=self.user_id).one()
+        self.user = User.query.filter_by(id=self.user_id).one()
```
ppanero, line 99: two comments:
- What happens if it does not exist? there is not exception `except` for `NoResultFound`.
- When querying the `User` do we get the `UserProfile` and `UserIdentity` as references or similar that can be used to avoid db queries?

site/cds_rdm/ldap/utils.py
```diff
@@ -0,0 +1,119 @@
...
+        self.data = self._get_full_user_info()
+
+    def _get_full_user_info(self):
```
ppanero, line 102: it seems a bit counter intuitive to have a function called in the constructor to denormalize data also instantiated in the constructor (i.e. `self.user_profile`).

ppanero (commented):
None of the reviewers (me and David) have worked with LDAP so our knowledge there is limited. Might need another review.

We did not have time to go through `api.py`, `client.py` and `user_importer.py`, those still need reviewing.

site/cds_rdm/ldap/errors.py
kpsherva: should go to the logs below

PR #91, opened 2023-07-21
landing-page: add conceptdoi

zzacharo:
Screenshot 2023-07-21 at 18 02 52

### DIFF
--- a/site/cds_rdm/assets/semantic-ui/js/cds_rdm/src/records/detail.js
+++ b/site/cds_rdm/assets/semantic-ui/js/cds_rdm/src/records/detail.js
@@ -1,0 +1,14 @@
+/*
+ * This file is part of Invenio.
+ * Copyright (C) 2016-2022 CERN.
+ *
+ * Invenio is free software; you can redistribute it and/or modify it
+ * under the terms of the MIT License; see LICENSE file for more details.
+ */
+
+import $ from "jquery";
+
+// Initialize conceptdoi modal
+$("#record-conceptdoi-badge").on("click", function () {
+  $("#conceptdoi-modal").modal("show");
+});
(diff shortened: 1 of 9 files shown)


site/cds_rdm/assets/semantic-ui/js/cds_rdm/src/records/detail.js
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (C) 2016-2022 CERN.
+ *
+ * Invenio is free software; you can redistribute it and/or modify it
+ * under the terms of the MIT License; see LICENSE file for more details.
+ */
+
+import $ from "jquery";
+
+// Initialize conceptdoi modal
+$("#record-conceptdoi-badge").on("click", function () {
```
kpsherva, line 12: shelve: ideally the event should be generic for any modal, not depending on the element ID

templates/semantic-ui/invenio_app_rdm/records/details/doi.html
```diff
@@ -0,0 +1,19 @@
+{#
+    Copyright (C) 2023 CERN.
+
+    CDS-RDM is free software; you can redistribute it and/or modify it
+    under the terms of the MIT License; see LICENSE file for more details.
+#}
+
```
kpsherva, line 7: minor: jinja formatting missing

kpsherva (approved):

site/cds_rdm/assets/semantic-ui/js/cds_rdm/src/records/detail.js
ntarocco: I am not sure that I understand your comment, we want to open that exact modal, no?

ntarocco (approved):
👏🏼

site/cds_rdm/assets/semantic-ui/js/cds_rdm/src/records/detail.js
kpsherva: yes, only the one specific modal should be opened, but it is possible to implement in a generic way (if not already done in the plain semantic)

PR #95, opened 2023-08-06
groups: fetch async to improve login performance

ntarocco:
Requires: https://github.com/inveniosoftware/invenio-oauthclient/pull/315

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -201,6 +201,7 @@ SECURITY_SEND_REGISTER_EMAIL = False
 # See https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/config.py
 from invenio_oauthclient.contrib.keycloak import KeycloakSettingsHelper
 from cds_rdm.oidc import (
+    cern_info_handler,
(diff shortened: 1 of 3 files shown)


ntarocco:
For reviewers (@zzacharo @kpsherva), given that we have remove the `UserNeed(current_user.email)` [here](https://github.com/inveniosoftware/invenio-oauthclient/pull/317), the [permission generator by user e-mail](https://github.com/CERNDocumentServer/cds-rdm/blob/master/site/cds_rdm/generators.py#L42) will not work anymore. We might want to remove it (and keep only groups), or explore alternatives.

zzacharo:
@ntarocco @kpsherva Far from ideal but if we really want to keep the feature "permission by email" we could replace the emails with user ids of these users...

ntarocco:
> @ntarocco @kpsherva Far from ideal but if we really want to keep the feature "permission by email" we could replace the emails with user ids of these users...

That would be easy to fix in the code, a bit less usable in the config (you need to find out the user id). Why not... up to you :)

PR #106, opened 2023-09-26
[TO BE MERGED WHEN CODE UPGRADE] details: add include_deleted to macros

jrcastro2:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1479


zzacharo:
template was removed after code upgrade, so this is not longer needed.

PR #168, opened 2024-07-04
Offloading

kpsherva:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/167

### DIFF
--- a/site/cds_rdm/files.py
+++ b/site/cds_rdm/files.py
@@ -11,21 +11,68 @@
 import mimetypes
 import os
 import unicodedata
-from urllib.parse import quote
+from urllib.parse import quote, urlsplit, urlunsplit
 
+import requests
 from flask import current_app, make_response, request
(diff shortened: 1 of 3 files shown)


site/cds_rdm/files.py
```diff
@@ -45,12 +92,18 @@ def send_file(
             and current_app.config["FILES_REST_XSENDFILE_ENABLED"]
         )
 
+        if not offload_enabled:
+            # don't offload if not enabled
+            return super().send_file(filename, **kwargs)
```
ntarocco, line 97: Nothing to do with your changes, but I just noticed that here we don't pass the explicit kwargs (mimetype, ...). Can you check the upper send_file if it expects them?

ntarocco (approved):

site/cds_rdm/files.py
kpsherva: done, added

Issue #142, opened 2024-03-28
Bootstrap migration


ntarocco:
- Decide how to organize/structure the code
  - use rdm-migrator?
  - use service layers when importing?
  - resume migrator kit/UI

### Done when
We are able to:
1. export from CDS
2. review transformation via UI
3. import in the new CDS

This excludes any transformation rule: having title and a fake creator is enough

kpsherva:
![Image](https://github.com/CERNDocumentServer/cds-rdm/assets/38131488/88afc18d-702f-4c2d-baee-650b25886971)

PR #169, opened 2024-07-05
migrator: add migration for SSPN

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/142


site/setup.cfg
```diff
@@ -10,6 +10,7 @@ python_requires = >=3.9
 zip_safe = False
 install_requires =
     python-ldap>=3.4.0,<3.5.0
+    #missing invenio-rdm-migrator?
```
zzacharo, line 13: it is missing because of the sqlalchemy conflict but you can add it back now when your fix is merged

site/setup.cfg
```diff
@@ -30,12 +31,6 @@ invenio_assets.webpack =
     cds_rdm_theme = cds_rdm.webpack:theme
 invenio_celery.tasks =
     cds_rdm = cds_rdm.tasks
-#flask.commands =
```
ntarocco, line 33: why commenting all this out, and above, and not simply removing it?

ntarocco (approved):

site/setup.cfg
kpsherva: sorry, missed this commented our piece
invenio-rdm-migrator will not be installed here at all

site/setup.cfg
kpsherva: sorry I am confused, this PR removes it, no?

PR #172, opened 2024-07-09
Archiver permissions policy

kpsherva:


zzacharo (approved):

ntarocco (approved):
LGTM! Thank you for adding the test!

site/cds_rdm/generators.py
```diff
@@ -46,3 +49,28 @@ def needs(self, **kwargs):
     def query_filter(self, **kwargs):
         """Match all in search."""
         raise NotImplementedError
+
+
+class AuthenticatedRegularUser(AuthenticatedUser):
```
zzacharo, line 54: I am not sure how I feel about the naming here but maybe we can think it another time...

Issue #193, opened 2024-09-02
names vocab: allow names vocab to have 2 types of objects


jrcastro2:
There will be 2 types of objects in the `names` vocab: the ORCID objects, and the CERN objects.

The ORCID objects are inserted or updated via the automatic jobs/dump. Its IDs cannot really be changed as all the code relies on having the ORCID as the id:
```
- affiliations:
  - name: Northwestern University
  family_name: Findler
  given_name: Robby
  id: 0000-0002-4245-2000
  identifiers:
  - identifier: 0000-0002-4245-2000
    scheme: orcid
```

The CERN objects **could look like this** (check `props` definition in other vocabs):
```
- affiliations:
  - name: <affiliations from CERN db>
  family_name: Findler
  given_name: Robby
  id: <encrypted-person-id>
  props:
    e-mail: <email>
    username: <username>
    is_cern: True
  identifiers:
  - identifier: <encrypted-person-id>
    scheme: CERN
```

In general, the `vocabularies` schemas and features should be enhanced to support:
1. extra fields for the `names`. It looks like that it is not the case today, see [here](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/contrib/names/jsonschemas/names/name-v1.0.0.json#L6).
2. vocabularies should support deprecation.

ntarocco:
At the end, we will need to update [this issue](https://github.com/CERNDocumentServer/cds-rdm/issues/135) with the latest implementation

PR #195, opened 2024-09-04
sync ldap into names

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/193

# As a non admin user, see the synced value
image

# As an admin, both values, the deprecated one appears greyed
image


site/cds_rdm/jobs.py
```diff
@@ -0,0 +1,53 @@
...
+    description = "Merge duplicate names in the names vocabulary."
+    task = merge_duplicate_names_vocabulary
+
+    @classmethod
+    def build_task_arguments(cls, job_obj, since=None, **kwargs):
+        """Build task arguments."""
+        if since is None and job_obj.last_runs["success"]:
+            since = job_obj.last_runs["success"].started_at
+        else:
+            since = (datetime.now() - timedelta(days=1)).isoformat()
```
kpsherva, line 49: shouldn't be needed, ping @ntarocco to confirm

site/cds_rdm/jobs.py
```diff
@@ -0,0 +1,53 @@
...
+    id = "sync_local_accounts_to_names"
+    title = "Sync local accounts to names"
+    description = "Sync local accounts to names vocabulary."
+    task = sync_local_accounts_to_names
+
+
+    @classmethod
+    def build_task_arguments(cls, job_obj, since=None, user_id=None, **kwargs):
+        """Build task arguments."""
+        if since is None and job_obj.last_runs["success"]:
```
kpsherva, line 28: shouldn't be needed

site/cds_rdm/jobs.py
jrcastro2: I thought only needed for the first run. However this has to be updated once jobs is refactored, might be solved by then

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
     except Exception as e:
         db.session.rollback()
         current_app.logger.exception(e)
+
+
+@shared_task()
+def sync_local_accounts_to_names(since, user_id=None):
+    """Syncs local accounts to names vocabulary.
+
+    If the name is marked as "unlisted" - meaning it is deprecated - we still update it with the new values, if needed.
```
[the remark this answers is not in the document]
jrcastro2: Right now we only have one use case where an user is considered deprecated. This happens when we have the user is coming from ORCID and coming from CERN database with the ORCID registered there as well, in these cases we mark the CERN user as deprecated and merge the extra data in the ORCID value.

However, there can be more cases in the future where we want to deprecate names, not sure about the one you mention - when a user leaves CERN - is a valid one, as per my understanding we might want to still add that person as contributor of a project although they have left. However it might be a valid case or there might be others.

Existing records are not affected by deprecated values, this only affects the search display,  the depreacted values won't be shown.

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+    prop_values = ["group", "department", "maibox", "section"]
+    service = current_service_registry.get("names")
+
+    def _check_if_update_needed(user, name, is_orcid=False):
+        """Check if the name needs to be updated."""
+
+        updated = False
+        updated_name = {**name}
+
+        if not is_orcid:
```
kpsherva, line 167: I guess this means orcid takes precedence as a source of truth when it comes to updates
I am not sure though if thats what we want to achieve, I think CERN will be more reliable than orcid to be up to date regarding name and surname, no?

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+            user_prop = user.user_profile.get(prop)
+            name_prop = name.get("props", {}).get(prop)
+            if user_prop and user_prop != name_prop:
+                if "props" not in updated_name:
+                    updated_name["props"] = {}
+                updated_name["props"][prop] = user_prop
+                updated = True
+
+        # add the email as props
+        if user.email and user.email != name.get("props", {}).get("email"):
```
kpsherva, line 208: how is email then consumed?

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+        try:
+            service.create(system_identity, name)
+        except ValidationError as e:
+            current_app.logger.error(f"Error creating name for user {user.id}: {e}")
+
+    users = []
+    # Allows to sync a single user
+    if user_id:
+        user = User.query.get(user_id)
+        users = [user] if user else []
```
kpsherva, line 250: I think
```suggestion
        users = [user]
```
won't `.get` raise an exeption if user is not found for the given id?

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+                    _create_new_name(user)
+            else:
+                _create_new_name(user)
+
+
+
+@shared_task()
+def merge_duplicate_names_vocabulary(since=None):
+    """Merges duplicate names in the names vocabulary."""
+    service = current_service_registry.get("names")
```
kpsherva, line 276: just for readability
```suggestion
    names_service = current_service_registry.get("names")
```

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+
+    def _merge(name_from, name_to):
+        """Merges the names.
+
+        param name_from: The name to merge from. This will marked as unlisted after the merge, meaning that won't be returned in the search results.
+        param name_to: The name to merge to.
+        """
+        updated = False
+        # Merge the affiliations
+        affiliations_to = name_to.get("affiliations", [])
```
kpsherva, line 286: it would be a little bit less "tangled" if we call it `affiliations_source` instead of `affiliations_from`  and `affiliations_dest`, but lets hear what the rest of the team thinks before changing it

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+        for affiliation_from in name_from.get("affiliations", []):
+            if affiliation_from not in affiliations_to:
+                affiliations_to.append(affiliation_from)
+                updated = True
```
kpsherva, line 290: sorry I am really lost on `_from` and `_to`, not sure which one is the "new" update

I think this could be a bit faster (I didn't check though!) to read from in terms of merging lists, no need to check each element, set ensures no duplication
```suggestion
        affiliations_from = name_from.get("affiliations", [])
        affiliations_to = name_to.get("affiliations", [])
        if affliations_from != affiliations_to:
            affiliations_to = list(set(affiliations_from + affiliations_to))
            updated = True        
```

but before changing lets check if the team agrees

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+    filters = [
+        dsl.Q("term", **{"props.is_cern": True}),
+        dsl.Q("term", **{"identifiers.scheme": "orcid"}),
+        dsl.Q("bool", must_not=[dsl.Q("term", tags="unlisted")]),
+    ]
+    if since:
+        filters.append(dsl.Q("range", updated={"gte": since}))
+    combined_filter = dsl.Q("bool", filter=filters)
+
+    names = service.scan(system_identity, extra_filter=combined_filter)
```
kpsherva, line 342: shouldn't we rely on the db table instead?

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+        )
+        if orcid:
+            names_to_merge = service.resolve(
+                system_identity, id_=orcid, id_type="orcid", many=True
+            )
+            if names_to_merge.total > 1:
+                # We get the ORCID value and merge all the other values into it (Ideally there should be only 1 value apart from the ORCID)
+                orcid_name = None
+                other_names = []
+                for name_to_merge in names_to_merge.hits:
```
kpsherva, line 361: sorry for a silly question, but I am not sure what condition we use to determine if we merge the name or not. Could we discuss this during the demo?

ntarocco: I think that is not super clear just because there is a lot of code.
We merge when the CERN user has an ORCID in her/his profile, and the ORCID is found in the names vocab. from the ORCID dump. Basically, there is an ORCID match.

kpsherva: maybe stating this in a comment in the code could help to understand whats the required condition to merge. I think it should be documented/destribed with more details

site/cds_rdm/jobs.py
jrcastro2: We have the same here: https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/jobs.py#L60-L63

But not sure I understand well, does this means that our funcs do need ot accept since=None?

site/cds_rdm/tasks.py
jrcastro2: The email is displayed in the UI 
image

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,224 @@ def sync_users():
...
+        except ValidationError as e:
+            current_app.logger.error(f"Error creating name for user {user.id}: {e}")
+
+    users = []
+    # Allows to sync a single user
+    if user_id:
+        user = User.query.get(user_id)
+        users = [user] if user else []
+    else:
+        users = User.query.filter(User.updated > since).all()
```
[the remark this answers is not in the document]
jrcastro2: We changed it on the job to set it to 1 day ago if None, we can update it as we desire.

site/cds_rdm/tasks.py
jrcastro2: Hmm this could work with lists that only contain values, but in this case the lists contain dicts in different forms. For instance
```python
affiliations = [{"name": <value>}, ....]
identifiers = [{"identifier": <value>, "scheme": "orcid"}]
```

This lists are not hashable therefore cannot use `set` on them.

However I will try to improve this to make ir more readable.

ntarocco:
For the `Dep - Group` label, shall we put it even small, inside the parenthesis just after the email? It kind of belongs to the CERN part:
`(ORCID logo ORCID ID      CERN logo email depgroup)`

site/cds_rdm/tasks.py
jrcastro2: Given that we are applying a lot of filters, plus there depending on the time range and/or when we just imported ORCID and the CERN authors, there might be a lot of results to process and there are millions of values in this table, I would expect this to be quite slow.

I am aware that the DB is the source of truth but mainly for performance optimization I would go with OS, however, if there are strong reasons to prioritize the DB, I can change it and test it with a bigger sample to asses the performance

site/cds_rdm/authors/schema.py
```diff
@@ -0,0 +1,9 @@
+from invenio_records_resources.services.records.schema import BaseRecordSchema
+from marshmallow import fields
+
+
+class AuthorSchema(BaseRecordSchema):
+    """Author schema."""
+
+    given_name = fields.Str(required=True)
+    family_name = fields.Str(required=True)
```
jrcastro2, line 9: Keep last name, first name, affiliations

site/cds_rdm/tasks.py
```diff
@@ -140,3 +149,495 @@ def sync_users():
...
+            try:
+                names_service.update(
+                    system_identity, name_dict.get("id"), updated_name, uow=uow
+                )
+            except ValidationError as e:
+                current_app.logger.error(f"Error updating name for user {user.id}: {e}")
+
+    def _update_author(name_id, user, uow=None):
+        """Updates the author record with the new values."""
+        author = current_authors_service.read(system_identity, name_id)
```
jrcastro2, line 343: review to see if this is correct

site/cds_rdm/tasks.py
```diff
@@ -140,3 +149,495 @@ def sync_users():
...
+                # 2. We update the author and name value if it exists
+                with UnitOfWork(db.session) as uow:
+                    _update_name(user, cds_name_dict, uow=uow)
+                    _update_author(cds_name_dict["id"], user, uow=uow)
+                    uow.commit()
+            else:
+                # 3. If the name doesn't exist, we create a new one
+                try:
+                    with UnitOfWork(db.session) as uow:
+                        author = _create_author(
```
jrcastro2, line 506: Add the user_id to author and change to create_or_get_author()

ntarocco (commented):
Added comments, but mainly cosmetic.

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,430 @@ def sync_users():
...
+
+    def _create_new_name(user, deprecate=False, uow=None):
+        """Creates a new name for the user.
+
+        param user: The user object.
+        """
+
+        default_props = _get_default_props(user.id)
+        name = {
+            "id": str(user.user_profile.get("person_id")),
```
ntarocco, line 345: we can't use `.get` if it is expected, required.

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,430 @@ def sync_users():
...
+            "props": default_props,
+            "identifiers": [
+                {"scheme": "cds", "identifier": str(user.user_profile.get("person_id"))}
+            ],
+        }
+
+        if user.user_profile.get("given_name"):
```
ntarocco, line 352: Maybe use yet-another utiliy here?
`add_if_exists(name, "given_name", user_profile.get("given_name"))`?

site/cds_rdm/schemes.py
```diff
@@ -57,10 +57,25 @@ def is_inspire(val):
...
         "url_generator": None,
     }
+
+def is_cds(val):
+    """Test if argument is a valid CDS person id."""
+    pattern = r"^\d+$"
+    return bool(re.match(pattern, val))
+
+
+def cds():
```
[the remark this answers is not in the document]
jrcastro2: Yes, we have to pass it to `idutils.custom_schemes` entrypoint so that it can detect the scheme when submitting the record.

site/cds_rdm/tasks.py
```diff
@@ -140,3 +147,430 @@ def sync_users():
...
+
+        param user: The user object.
+        """
+
+        default_props = _get_default_props(user.id)
+        name = {
+            "id": str(user.user_profile.get("person_id")),
+            "props": default_props,
+            "identifiers": [
+                {"scheme": "cds", "identifier": str(user.user_profile.get("person_id"))}
```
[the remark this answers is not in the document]
jrcastro2: We filter out the ones that do not have `person_id`. At this point all values do have `person_id`

site/cds_rdm/tasks.py
jrcastro2: Hmm we could definitely do it but, it would be a complex one if we want to cover all the cases as we do not always assing the value directly, sometimes we build a dict as for the affiliations, for others like affiliations we want to append and for the props they are nested. I thought it would simpler like this but I am happy to update it if required.

PR #199, opened 2024-09-19
global: add redirection from legacy recids

kpsherva:


site/cds_rdm/resolver.py
```diff
@@ -0,0 +1,14 @@
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_pidstore.resolver import Resolver
+
+
+def get_pid_by_legacy_recid(pid_value):
```
carlinmack, line 5: nitpick: could the argument be `legacy_pid_value` as there's visual repetition of `pid_value` in the function which could be confusing

carlinmack (commented):
Other than the failing tests, LGTM + @0einstein0 🐞

site/cds_rdm/resolver.py
```diff
@@ -0,0 +1,14 @@
...
+    """Get ils record by pid value and pid type."""
```
zzacharo, line 6:
```suggestion
    """Get CDS record by pid value and pid type."""
```

zzacharo (approved):

Issue #183, opened 2024-08-02
legacy: migrated files redirection


kpsherva:
> The old URL path should still work and directly link to the file and versions should redirect correctly to the new one.
Double-check all possible URLs path/routes. It looks like that [there is some code done here](https://gitlab.cern.ch/cds-team/cds-legacy/blob/master/src/wn-cdsweb/lib/python/invenio/bibdocfile_webinterface.py#L184), but it might not take into account URL params (and we should).

from https://github.com/CERNDocumentServer/cds-rdm/issues/129

1. Find all the possible instances of `MIGRATED` in the current CDS, understand how it works
2. prepare code modifications to redirect MIGRATED records also with the entire subpath/query string. Be aware that redirection is already in place for CDS Videos and CLC, and we need to test that we don't break those after the modifications.
3. in CDS-RDM, implement something similar to [Zenodo redirector](https://github.com/zenodo/zenodo-rdm/blob/master/site/zenodo_rdm/redirector.py) to re-map the incoming redirection from current CDS, transform the URL, and re-redirect the final place. Example of possible final URL [this](https://dev-cds-rdm.web.cern.ch/records/ccgjf-zh606/preview/Pablo_arrutia_Report.pdf).

We need to take into account the multiple versions of a file.

Have a look to the [migration code](https://github.com/CERNDocumentServer/cds-migrator-kit/) in cds-migrator-kit.

PR #222, opened 2024-10-23
views: Add files redirection for legacy recids

sakshamarora1:
closes: #183


site/cds_rdm/views.py
```diff
@@ -8,16 +8,34 @@
...
+    if filename:
+        url_path = url_for(
+            "invenio_app_rdm_records.record_file_preview",
+            pid_value=pid,
+            filename=filename,
+            **request.args,  # Transform if required
+        )
+    else:
+        url_path = record_url_for(pid_value=pid)
+    current_app.logger.debug(url_path)
```
zzacharo, line 38: Is this intentional?

site/cds_rdm/views.py
```diff
@@ -8,16 +8,34 @@
...
     """Redirect legacy recid."""
-    pid = get_pid_by_legacy_recid(id)
-    url_path = record_url_for(pid_value=pid)
+    pid = get_pid_by_legacy_recid(legacy_id)
+    if filename:
+        url_path = url_for(
+            "invenio_app_rdm_records.record_file_preview",
+            pid_value=pid,
+            filename=filename,
+            **request.args,  # Transform if required
```
zzacharo, line 34: Have you checked the legacy URL parameters that were supported in the files API?

site/tests/test_views.py
```diff
@@ -0,0 +1,71 @@
...
+
+    rdm_record_url = record_url_for(pid_value=record.id)
+
+    # Test record redirection
+    response = client.get("/record/123456")
+    assert response.status_code == 302
+    assert rdm_record_url in response.location
+
+    response = client.get("/record/123456?test=check&foo=bar")
+    assert response.status_code == 302
```
zzacharo, line 57: can you also check the redirected response?

jrcastro2 (commented):
LGTU! Just a comment about the non-exisitng version. Once this is addressed it looks good to us. Peer reviewed with Fatimah

site/cds_rdm/resolver.py
```diff
@@ -8,16 +8,41 @@
...
+        return latest_record
+
+    # Use the version number to get the desired record pid value
+    record = current_rdm_records_service.search_versions(
+        identity=g.identity,
+        id_=latest_record["id"],
+        extra_filter=dsl.Q("term", **{"versions.index": version}),
+    ).to_dict()["hits"]["hits"]
+    if not record:
+        return latest_record
```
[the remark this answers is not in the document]
sakshamarora1: Using [this function](https://github.com/inveniosoftware/invenio-drafts-resources/blob/a5bb49b5dc00830e050207d919ab14ea1a3a47a3/invenio_drafts_resources/services/records/service.py#L150)
If we raise and handle exception in the linked function, it wouldn't change how we handle the record not being found here. If we raise an exception there, and handle it here, then we'll also have to handle the exception everywhere this function is being used, also it will error out for the api. Or maybe I missed something here? Please let me know 😅

site/cds_rdm/resolver.py
```diff
@@ -8,16 +8,41 @@
...
+    # Use the object uuid to get the parent pid value
+    parent_pid = PersistentIdentifier.query.filter_by(
+        object_uuid=recid.object_uuid, pid_type="recid"
     ).one()
-    return pid.pid_value
+
+    return parent_pid.pid_value
+
+
+def get_record_by_version(parent_pid_value, version):
```
kpsherva, line 32: didn't we agree to add this to API class?

site/cds_rdm/resolver.py
```diff
@@ -8,16 +8,41 @@
...
+    """Get record by parent pid value and version."""
+    latest_record = current_rdm_records_service.read_latest(
+        identity=g.identity, id_=parent_pid_value
+    )
+    if not version or version == "all" or latest_record["versions"]["index"] == version:
+        return latest_record
+
+    # Use the version number to get the desired record pid value
+    record = current_rdm_records_service.search_versions(
```
kpsherva, line 41: I think retrieving versions from search is flaky, I would suggest to use record api

site/cds_rdm/views.py
```diff
@@ -8,19 +8,46 @@
...
-    pid = get_pid_by_legacy_recid(id)
+    pid = get_pid_by_legacy_recid(legacy_id)
     url_path = record_url_for(pid_value=pid)
     return redirect(url_path)
 
 
+def legacy_files_redirect(legacy_id, filename):
+    """Redirection for legacy files."""
+    parent_pid = get_pid_by_legacy_recid(legacy_id)
+    query_params = request.args.copy()
```
kpsherva, line 35: nit: it would be a bit cleaner to wrap this in a decorator (check invenio-app-rdm -> records_ui->views->decorators.py)

site/cds_rdm/resolver.py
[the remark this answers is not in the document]
kpsherva: ping @ntarocco how about having specific 404 page with a link pointing to lastest version ? in case of the specific requested one was not there (deleted). I agree with Javi that redirecting by default to the latest, when you requested a specific version in the URL, without any warning, might be confusing for the user.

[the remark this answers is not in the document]
ntarocco: We discussed IRL with Saksham and decided to go with the latest version.
I agree that a 404 with a link to the latest might be better indeed.
However, it depends on how hard it is. @sakshamarora1 can you check? If it takes more than half a day, let's go with the current implementation.

site/cds_rdm/views.py
ntarocco: Why a decorator? This is used only in one specific view.

site/cds_rdm/views.py
```diff
@@ -8,19 +8,46 @@
...
+    version = query_params.pop("version", None)
+    record = get_record_by_version(parent_pid, version)
+    # Directly download files from redirected link to replicate the `allfiles-` behaviour from legacy
+    if filename.startswith("allfiles-"):
+        url_path = record["links"]["archive"]
```
ntarocco, line 40: Nicely done! :)

site/cds_rdm/views.py
```diff
@@ -8,19 +8,46 @@
 
 """Additional views."""
 
-from flask import Blueprint, redirect
```
ntarocco, line 11: I propose to move this a specific migration or redirector module/file as we did in Zenodo [see here](https://github.com/zenodo/zenodo-rdm/blob/master/site/zenodo_rdm/redirector.py).
In this way it will be more clear that this is needed to redirect legacy URLs.

ntarocco (commented):
Reviewed by Nico and Carlin with ❤️

site/cds_rdm/views.py
zzacharo: If we go with something like migration folder, we could move also the [models](https://github.com/CERNDocumentServer/cds-rdm/blob/master/site/cds_rdm/models.py) from here :)

sakshamarora1: Updated the code with the suggestions :D

site/cds_rdm/resolver.py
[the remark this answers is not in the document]
sakshamarora1: Added the template!

site/cds_rdm/migration/resolver.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
```
kpsherva, line 1: last comment about the folder naming. I would maybe propose "legacy" or "redirect" instead of "migration", I would like to avoid confusion with the actual migration code (even if not in this repo)

zzacharo (approved):

zzacharo (approved):

site/cds_rdm/legacy/resolver.py
```diff
@@ -0,0 +1,51 @@
...
+def get_pid_by_legacy_recid(legacy_recid):
+    """Get record by pid value."""
+    # Get the object uuid from pidstore
+    recid = PersistentIdentifier.query.filter_by(
+        pid_value=legacy_recid, pid_type="lrecid"
+    ).one()
+
+    # Use the object uuid to get the parent pid value
+    parent_pid = PersistentIdentifier.query.filter_by(
+        object_uuid=recid.object_uuid, pid_type="recid"
```
slint, line 28:
```suggestion
        object_uuid=recid.object_uuid, object_type="rec", pid_type="recid"
```

minor: otherwise you're not using the [(`object_type`, `object_uuid`) index](https://github.com/inveniosoftware/invenio-pidstore/blob/master/invenio_pidstore/models.py#L100), which will become pretty slow once the PIDs table gets big 😅

slint (approved):
One minor comment on using the right table index for PIDs lookup.

PR #223, opened 2024-10-23
Pipfile: Install jsonschema to fix RefReslutonError

sakshamarora1:
To fix tests:

```
ERROR tests/test_permissions.py::test_archiver_permissions - jsonschema.exceptions._RefResolutionError: <urlopen error unknown url type: local>
```


zzacharo (commented):
You need to relock the Pipfile but otherwise it looks good!.Thanks for the fix!

zzacharo (approved):

PR #224, opened 2024-10-23
cli: add commands to create and delete read only users

zzacharo:


zzacharo:
we will not need that

PR #241, opened 2024-11-01
assets: move custom fields into basic info section

kpsherva:


slint (approved):

assets/js/components/deposit/BasicInformation.js
```diff
@@ -0,0 +1,27 @@
...
+
+export class BasicCERNInformation extends Component {
+  render() {
+    const { customFieldsUI, record } = this.props;
+    return (
+      <CustomFields
+        config={customFieldsUI}
+        record={record}
+        templateLoaders={[
+          (widget) => import(`@templates/custom_fields/${widget}.js`),
```
ntarocco, line 13: I was not aware of this, and I find it dangerous, as there is no build time error if the path does not exist.

This will bite us back one day.

PR #255, opened 2024-11-15
stats: add explicit filter of robots needed for migrated events

zzacharo:


invenio.cfg
```diff
@@ -550,3 +550,14 @@ VOCABULARIES_NAMES_SCHEMES = {
     "lcds": {"label": _("CDS"), "validator": schemes.is_legacy_cds, "datacite": "CDS"},
 }
 """Names allowed identifier schemes."""
+
+
+# Invenio Stats
+# =============
+
+# We need to add explicitely the `filter_robots` as the migrated legacy statistics are
+# were not being processed through the event preprocessors
```
ntarocco, line 559: Improve comment? are/were? and where did we add the `filter_robots`?

site/cds_rdm/stats/templates/events/file_download/os-v2/file-download-v1.0.0.json
```diff
@@ -0,0 +1,104 @@
...
+      "via_api": {
+        "type": "boolean"
+      },
+      "is_lcds": {
+        "type": "boolean",
+        "description": "This field marks all statistical events that have been migrated from the legacy CDS system."
+      },
+      "before_COUNTER": {
+        "type": "boolean",
+        "description": "This field applies to all migrated events where no information was available to determine whether they were valid or robot events. This was later resolved with the implementation of a proper robot-checking mechanism, ensuring COUNTER compliance."
```
ntarocco, line 94: Maybe human instead of valid?
```suggestion
        "description": "This field applies to all migrated events where no information was available to determine whether they were human or robot events. This was later resolved with the implementation of a proper robot-checking mechanism, ensuring COUNTER compliance."
```

ntarocco (commented):
will you add the config to have yearly indices?

invenio.cfg
zzacharo: leftover I will remove it

zzacharo:
> will you add the config to have yearly indices?

If you agree, I would do this in a separate issue as it needs some testing.

invenio.cfg
```diff
@@ -550,3 +550,11 @@ VOCABULARIES_NAMES_SCHEMES = {
     "lcds": {"label": _("CDS"), "validator": schemes.is_legacy_cds, "datacite": "CDS"},
 }
 """Names allowed identifier schemes."""
+
+
+# Invenio Stats
+# =============
+
+# We override the templates to add new fields needed for the migrated statistic events
+_APP_RDM_STATS_EVENTS["file-download"]["templates"] = "cds_rdm.stats.templates.events.file_download"
```
kpsherva, line 559: I guess we will need to remember on the deployment too? should we note it down somewhere?

zzacharo: you mean update of the mapping? I will do it yes

kpsherva (approved):

Issue #213, opened 2024-10-15
URL redirection


ntarocco:
Implement URL redirection of:
* `/collection/CERN%20Students%20Projects` -> `/communties/<SSPN>`
* `/search?...cc=CERN Students Projects...[p=<query>]` -> `/communties/<SSPN>/search?q=<query>`
* ~~Also with `c` only on not? Check `search_engine.py`~~ not for now
* ~~Ideally, we should parse the `p=` param, to detect if there is a Marc21 field, and redirect to the same search re-mapped. Example: `p=245__a:my title` (or `p=245:my title`) should redirect to `q=metadata.title:my title`. The issue is that some fields change per collection. Do we take it into account?~~ Not for now
* Redirect submission form for sspn, when url is https://cds.cern.ch/submit and has parameters `comboDEPIN=CSPN` and `doctype=DEPIN`, redirect to https://dev-cds-rdm.web.cern.ch/uploads/new?community=sspn

### Technical details

We don't want to expose a `/legacy` anywhere, therefore we can't respond with a `Redirect /legacy` from current CDS.

For the migrated collection:
1. update CDS webserver to redirect migrated paths `/<path>` to `repository.cern/<path>`. For example, `/collection/SSPN` to `repository.cern/collection/SSPN` 
2. in new CDS webserver, we should add an *internal* rewrite/redirection/rule in the main `location` similar to:

    when the host is not `repository.cern`
    AND
    the URL path matches one of the migrated paths (e.g. /collection, /files, /record)
    THEN
    add an internal rewrite to `/legacy/<path>`
    OTHERWISE
    the request should go to the `/` (maybe a valid URL for InvenioRDM.

PR #261, opened 2024-11-19
legacy: Add redirection for collections

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-rdm/issues/213#issue-2588412164


site/cds_rdm/legacy/redirector.py
```diff
@@ -59,19 +63,66 @@ def legacy_files_redirect(legacy_id, filename):
     return redirect(url_path)
 
 
+def legacy_collection_redirect(collection_name):
+    """Redirection for legacy collections."""
+    cds_community_uuid = current_app.config.get(
+        "CDS_REDIRECTION_COLLECTIONS_MAPPING", {}
+    ).get(collection_name, None)
```
ntarocco, line 70:
```suggestion
    cds_community_uuid = current_app.config["CDS_REDIRECTION_COLLECTIONS_MAPPING"].get(collection_name, None)
```

site/cds_rdm/legacy/redirector.py
```diff
@@ -59,19 +63,66 @@ def legacy_files_redirect(legacy_id, filename):
...
+    if not cds_community_uuid:
+        raise NoResultFound
+    community_data = get_community_by_uuid(cds_community_uuid)
```
ntarocco, line 73: I guess this will raise if not found? A mistake might happen here.

site/cds_rdm/legacy/redirector.py
```diff
@@ -59,19 +63,66 @@ def legacy_files_redirect(legacy_id, filename):
...
+        url_path = url_for(
+            "invenio_app_rdm_communities.communities_home",
+            pid_value=community_data.slug,
+            **query_params,
+        )
+    return redirect(url_path)
+
+
+def legacy_search_redirect():
+    """Redirection for legacy search."""
```
ntarocco, line 92: Can you be a bit more verbose with the docstring? It would be nice to read what is handling, and with one or more examples. E.g.
```
c=<colleciton name> -> /search?...
```

site/tests/legacy/test_redirector.py
```diff
@@ -106,24 +110,58 @@ def test_legacy_redirection(uploader, client, minimal_record_with_files, add_pid
...
     assert response.status_code == 404
 
     # files download redirection case
-    response = client.get("/record/123456/files/allfiles-small" + query_params)
+    response = client.get("/legacy/record/123456/files/allfiles-small" + query_params)
     assert response.status_code == 302
     assert response.location == record_v2.links["archive"]
+
+
+def test_legacy_collection_redirection(
```
ntarocco, line 136: Thank you for adding the tests!

site/cds_rdm/legacy/resolver.py
```diff
@@ -49,3 +53,12 @@ def get_record_by_version(parent_pid_value, version):
         # If record is not found, that means the version doesn't exist
         raise VersionNotFound(version=version, latest_record=latest_record)
     return hits[0]
+
+
+def get_community_by_uuid(community_uuid):
```
ntarocco, line 58: This is not needed and we should rely on the db instead.
You need to use [the read service method](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/communities/services/service.py#L664), see an [example here](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/communities/resources/resource.py#L119).

The identity should be system process but... pay attention that the community might be restricted. So we cannot leak any info. I would add a comment.
A redirection works because the permissions will be checked.

sakshamarora1: Okay! Why do we want to keep it `system_identity`?

sakshamarora1:
Tests will fail until https://github.com/inveniosoftware/invenio-app-rdm/pull/2923 is deployed. Otherwise will need to revert last commit (use g.identity and duplicate error handlers)
cc @zzacharo

site/cds_rdm/legacy/redirector.py
```diff
@@ -9,16 +9,18 @@
 """Redirector functions and rules."""
 
 from flask import Blueprint, current_app, redirect, render_template, request, url_for
+from invenio_communities.views.ui import not_found_error
 from invenio_rdm_records.resources.urls import record_url_for
+
+# from invenio_records_resources.services.errors import RecordPermissionDeniedError
```
kpsherva, line 15: leftover?

invenio.cfg
```diff
@@ -277,6 +277,11 @@ CDS_EOS_OFFLOAD_X509_KEY_PATH = ""
 # check nginx config for more details
 CDS_EOS_OFFLOAD_REDIRECT_BASE_PATH = ""
 
+# CDS Migration
+CDS_REDIRECTION_COLLECTIONS_MAPPING = {
+    "CERN Student Project Notes": "", # uuid
```
kpsherva, line 282: is it set on openshift? or we expect to set it when we have the ready uuid?

site/cds_rdm/legacy/resolver.py
kpsherva: because we won't be able to read from the service to obtain the slug. By the way, why not returning the slug only? it would be safer than returning the whole community item

site/cds_rdm/legacy/resolver.py
```diff
@@ -49,3 +51,12 @@ def get_record_by_version(parent_pid_value, version):
...
+    """Get community by uuid."""
+    community = current_communities.service.read(
+        id_=community_uuid,
+        identity=system_identity,  # using system_identity here as the permission check on user identity will be performed by the read service method
```
kpsherva, line 60:
```suggestion
        identity=system_identity,  # using system_identity here to retrieve the community slug, real permission check on actual user identity will happen in the view after the redirection
```

site/cds_rdm/legacy/redirector.py
```diff
@@ -59,19 +61,72 @@ def legacy_files_redirect(legacy_id, filename):
...
+    if query_params["q"]:
+        url_path = url_for(
+            "invenio_app_rdm_communities.communities_detail",
+            pid_value=community.data["slug"],
+            **query_params,
+        )
+    else:
+        url_path = url_for(
+            "invenio_app_rdm_communities.communities_home",
+            pid_value=community.data["slug"],
```
kpsherva, line 83: did you check if we can resolve this url (and the one above) using community uuid? we would be able to avoid creating a new method `get_community_by_uuid` and using system identity in there

zzacharo:
@sakshamarora1 can you update the pipfile with the released https://github.com/inveniosoftware/invenio-app-rdm/releases/tag/v13.0.0b1.dev19 ?

invenio.cfg
sakshamarora1: We should set it when we have it ready

zzacharo: Btw the collection name in the URL is `CERN Students Projects`. Can you fix that?

site/cds_rdm/legacy/redirector.py
```diff
@@ -59,19 +55,64 @@ def legacy_files_redirect(legacy_id, filename):
...
+    # If not, then fetch from collection list, URLs with only single 'c' will be redirected for now
+    if not collection_name:
+        collection_name = query_params.pop("c", None)
+    if not collection_name:
+        raise not_found_error()
+    # Add logic for other redirections from search params when we get there
+    url = url_for(
+        "cds_rdm.legacy_collection_redirect",
+        collection_name=collection_name,
+        **query_params,
```
zzacharo, line 95: Does all of the query params coming from legacy make sense here? For example, the `p` parameter is used to store the query string.

zzacharo (approved):
A small comment about query params on collections but I am ok to create a separate issue for that.

site/cds_rdm/legacy/redirector.py
sakshamarora1: I think it does, because it is possible have a mix of old and new query params in the link.

site/cds_rdm/legacy/redirector.py
```diff
@@ -59,19 +55,62 @@ def legacy_files_redirect(legacy_id, filename):
...
+    """
+    Redirection for legacy search. Transforms the legacy URL syntax into RDM URL syntax.
+
+        /legacy?cc=<legacy collection name>... -> /communities/<rdm_community_id>/records?...
+        /legacy?c=<legacy collection name>... -> /communities/<rdm_community_id>/records?...
+        /legacy?c=<legacy collection name>&p=<query>... -> /communities/<rdm_community_id>/records?q=<query>...
+    """
+    # Fetch current collection if it exists
+    collection_name = request.args.get("cc", None)
+    # If not, then fetch from collection list, URLs with only single 'c' will be redirected for now
```
ntarocco, line 83: can you make this comment more clear please?

site/cds_rdm/legacy/redirector.py
```diff
@@ -59,19 +55,62 @@ def legacy_files_redirect(legacy_id, filename):
...
+    if not collection_name:
+        collection_name = request.args.get("c", None)
```
ntarocco, line 85: `c` can be a list, are we getting the first?

sakshamarora1: YEs, we will only be redirecting URLs with a single c. I am setting this condition in the legacy apache config, so if there are multiple 'c' params, it won't get redirected here anyways.

PR #279, opened 2024-12-06
docker: prepend /legacy to requests from cds.cern.ch

sakshamarora1:
Fixes: #213

### DIFF
--- a/docker-services.yml
+++ b/docker-services.yml
@@ -19,6 +19,8 @@ services:
       - "INVENIO_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://cds-rdm:cds-rdm@db/cds-rdm"
       - "INVENIO_WSGI_PROXIES=2"
       - "INVENIO_RATELIMIT_STORAGE_URI=redis://cache:6379/3"
+      - "INVENIO_COMMUNITIES_IDENTITIES_CACHE_REDIS_URL=redis://cache:6379/4"
+      - "INVENIO_IIIF_CACHE_REDIS_URL=redis://cache:6379/0"
   frontend:
     build: ./docker/nginx/
     image: cds-rdm-frontend
--- a/docker/nginx/conf.d/common.conf
+++ b/docker/nginx/conf.d/common.conf
@@ -1,0 +1,12 @@
+include uwsgi_params;
+uwsgi_buffering off;
+uwsgi_request_buffering off;
+uwsgi_param Host $host;
+uwsgi_param X-Forwarded-For $proxy_add_x_forwarded_for;
+uwsgi_param X-Forwarded-Proto $scheme;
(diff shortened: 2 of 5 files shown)


docker-services.yml
```diff
@@ -19,6 +19,8 @@ services:
       - "INVENIO_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://cds-rdm:cds-rdm@db/cds-rdm"
       - "INVENIO_WSGI_PROXIES=2"
       - "INVENIO_RATELIMIT_STORAGE_URI=redis://cache:6379/3"
+      - "INVENIO_COMMUNITIES_IDENTITIES_CACHE_REDIS_URL=redis://cache:6379/4"
```
ntarocco, line 22: To be added to [cookiecutter](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/docker-services.yml#L25) if not already there.

docker/nginx/conf.d/common.conf
```diff
@@ -0,0 +1,12 @@
+include uwsgi_params;
```
ntarocco, line 1: This is nice! We should check if we can avoid having an extra different config file, to make it easier to deploy this.

docker/nginx/conf.d/default.conf
```diff
@@ -58,40 +58,30 @@ server {
   # in case of chunked transfer encoding
   proxy_http_version 1.1;
 
+  # Redirect requests from `cds.cern.ch` and prefix the path with `/legacy`
+  location ~ ^/(record/|collection/|search) {
+    if ($host != "repository.cern") {
+      rewrite ^/(.*)$ /legacy$request_uri;
+    }
+    # Pass request to the web-ui
```
ntarocco, line 66: is it actually an `else` or not? 
Not clear why the `/search` is not working, we need to debug more....

sakshamarora1: It's gonna happen regardless of the if condition, so for example if the request is coming from the the normal rdm `/search` endpoint, it will be passed to this without rewrite as the host name would be repository.cern, and if from legacy, it will get written to `/legacy/search`. That is what I intend to achieve here.

docker/nginx/conf.d/default.conf
```diff
@@ -11,6 +11,13 @@ upstream api_server {
   server web-api:5000 fail_timeout=0;
 }
 
+map $host $rewrite_to_legacy {
+  "repository.cern" '';
+  "new-cds.cern.ch" '';
+  "dev-cds-rdm.cern.ch" '';
```
ntarocco, line 17: only keep localhost host here?

ntarocco (approved):

Issue #284, opened 2024-12-13
Names: drop usage of `internal_id`


ntarocco:
After long discussions, the `internal_id` is actually not needed. We can keep it empty, to avoid creating yet another Alembic recipe to drop it.

Apply the needed changes to reflect this implementation. In the `names` vocabulary, we will have:
1. the full ORCiD dump, where the primary key is the ORCiD
2. the full CERN users db dump, where the primary key is the Person ID
3. when a CERN user also has an ORCiD, the ORCiD name will be enriched with extra CERN information, and the CERN name will be `unlisted` so that it won't be visible.

PR #300, opened 2024-12-19
names: remove internal id from sync

jrcastro2:
* Always sync CERN name, if orcid is present tag CERN name as unlisted
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/284

### DIFF
--- a/site/cds_rdm/tasks.py
+++ b/site/cds_rdm/tasks.py
@@ -113,7 +113,7 @@ def sync_local_accounts_to_names(since=None, user_id=None):
         orcid = user.user_profile.get("orcid")
         try:
             # 1. Prefer ORCID: if the user has an ORCID, we update the ORCID name with the
-            # CERN user info, and we unlist any previous CERN-only name
+            # CERN user info, and we also update and unlist any previous CERN-only name
             current_app.logger.debug(
                 f"Names sync | Fetching ORCID name for user {user.id}."
             )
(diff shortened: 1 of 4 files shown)


site/cds_rdm/tasks.py
```diff
@@ -125,7 +125,11 @@ def sync_local_accounts_to_names(since=None, user_id=None):
                         f"Names sync | Unlisting CERN name for user {user.id}."
                     )
                     names_utils.update_name(user, cern_name, unlist=True, uow=uow)
-
+                else:
+                    current_app.logger.debug(
+                        f"Names sync | No CERN name found for user {user.id}. Creating new unlisted name."
```
kpsherva, line 130: sorry I am a bit confused by the message. What is the use case for creating unlisted CERN name? is it when we have orcid entry but no related CERN user was found? why to create a CERN name at all?
how I understand the case when it is not needed to create unlisted:
1. I get an ORCID entry (from the dump). Most likely not a CERN researcher
2. I search for CERN user, I find none, because it is not a CERN person
3. I create a CERN name anyway, unlisted?

please let if I understood incorrectly

jrcastro2: Based on the discussion we had (we didn't specify this concrete case in the [issue](https://github.com/CERNDocumentServer/cds-rdm/issues/284) but the third point more or less describes it), I remember that we agreed to remove the person id from the `internal_id` but we keep the name when syncing it with ORCID value to have some sort of "history" and I guess it will be required for the migration, as we will match the unlisted value and not the ORCID value since there is no person id there anymore.

From now on, every time we sync if we find the ORCID value we create the unlisted CERN value as well, before it was simply merging to the ORCID value.

kpsherva: ok I think I understand now.
My initial wrong assumption was that ORCID dump is the source of truth, but now I realise you iterate on user profiles. Makes sense now.
Can you add a comment above this log message, explaining why we do this unlisted name creation?

ntarocco (approved):

PR #311, opened 2025-01-08
views: register custom index page

kpsherva:
* assets: fix responsive views, reorganise content
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/204

PR #313, opened 2025-01-10
Lcds scheme

kpsherva:
screenshots in https://github.com/CERNDocumentServer/cds-rdm/pull/311

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -503,7 +503,7 @@ RDM_RECORDS_IDENTIFIERS_SCHEMES = {**RDM_RECORDS_IDENTIFIERS_SCHEMES,
                                                   "datacite": "INSPIRE"}}}
 
 RDM_RECORDS_PERSONORG_SCHEMES = {**RDM_RECORDS_PERSONORG_SCHEMES,
-                                 **{"inspire": {"label": _("Inspire"),
+                                 **{"inspire_author": {"label": _("Inspire"),
                                                 "validator": schemes.is_inspire_author,
                                                 "datacite": "INSPIRE"},
                                     "lcds": {"label": _("CDS"),
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -144,7 +144,6 @@ THEME_SITENAME = 'CDS'
 # Site tracking code template for matomo analytics
 # Enable the below on the dedicated environment
 THEME_TRACKINGCODE_TEMPLATE = "cds_rdm/matomo-test.html"
-RDM_FRONTPAGE = False
 
 # Invenio-App-RDM
 # ===============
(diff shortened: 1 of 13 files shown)


site/cds_rdm/views.py
```diff
@@ -0,0 +1,47 @@
...
+        template_folder="templates",
+        static_folder="static",
+    )
+    blueprint.add_url_rule("/", "index", view_func=index)
+    return blueprint
+
+
+def index():
+    """Frontpage."""
+    featured_communities_search = current_communities.service.featured_search(
```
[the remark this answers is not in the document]
kpsherva: It is because the next step was to have "My communities" tab in this component, which would then be user-specific (check below)

invenio.cfg
```diff
@@ -144,6 +144,7 @@ THEME_SITENAME = 'CDS'
 # Site tracking code template for matomo analytics
 # Enable the below on the dedicated environment
 THEME_TRACKINGCODE_TEMPLATE = "cds_rdm/matomo-test.html"
+RDM_FRONTPAGE = False
```
slint, line 147: I see on Zenodo that we override the frontpage view, by [using the 2-tuple format for tha `APP_RDM_ROUTES` config](https://github.com/zenodo/zenodo-rdm/blob/ea0ef69fee51e8ab60a45bbfde3ad16d5d14976f/invenio.cfg#L111), which still allows to completely override the view function, i.e.:

```python
APP_RDM_ROUTES["index"] = ("/", frontpage_view_function)
```

This helps also with URL generation, since otherwise you can't use `url_for` to generate links to the frontpage (unless hardcoded to `"/"`, which wouldn't be great).

I remember, we struggled to find this approach and accidentally discovered it on some Discord thread at the RDM workshop last year 😅. To be honest, I think the `APP_RDM_ROUTES` config format is the wrong approach for defining these "common" RDM routes, since it also doesn't allow you to pass freely additional args to `Blueprint.add_url_rule(...)`.

utnapischtim: as a footnote:
[this](https://github.com/inveniosoftware/invenio-app-rdm/blob/50ec8ebbfdb80cc9586c15f34cee33f4ce7084ca/invenio_app_rdm/theme/views.py#L25) code enables that feature. which is limited to `APP_RDM_ROUTES` and `COMMUNITIES_ROUTES` in `invenio_app_rdm/theme/views.py` and `invenio_app_rdm/records_ui/views/__init__.py`

kpsherva, line 147:
```suggestion
```

site/cds_rdm/views.py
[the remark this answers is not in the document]
jrcastro2: Addressed by using Anonymous identity for the featured communities while keeping the user identity for my communities

jrcastro2 (approved):
LGTM!

Issue #256, opened 2024-11-15
Convert statistics to yearly


zzacharo:
- change this one: https://github.com/inveniosoftware/invenio-stats/blob/d534cbfc11b491c7438774aa6f8b0d6aee18c15c/invenio_stats/bookmark.py#L18
- change suffix [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/config.py#L1169) as [here](https://github.com/inveniosoftware/invenio-stats/commit/9e712e45979ba14bfae3f55698539127bdfc7e4d#diff-bad6dd9cbc5ddb20f9fa5050b6c380b4963a6155d47f41847f198d54171cf243R143) for both records and files

jrcastro2:
Deployment notes:

Converting to yearly stats
====

1. Run the CLI to process the events, so that it creates the indexes for the new year (ALTERNATIVELY we can simply create the indexes as done in the next steps)
    - `invenio stats events process file-download`
    - `invenio stats aggregations process file-download-agg`
    - `invenio stats events process record-download`
    - `invenio stats aggregations process record-download-agg`
2. For the old indexes, we need to create the new yearly ones and reindex the data.

Simply get the mapping to copy it over for the yearly index.
```
GET cds-rdm-events-stats-record-view-2024-XX/_mapping 
```

Create the yearly index with the previous mapping:
```
PUT cds-rdm-events-stats-record-view-2024
{
...
}

```

3. Deploy the changes


4. Reindex the data for all the years:
```
POST _reindex
{
  "source": {
    "index": "cds-rdm-events-stats-record-view-2024-*"
  },
  "dest": {
    "index": "cds-rdm-events-stats-record-view-2024"
  }
}
```

5. Rename the old monthly indexes (to be done for all months)
```
POST _reindex
{
  "source": {
    "index": "cds-rdm-events-stats-record-view-2024-01"
  },
  "dest": {
    "index": "deprecated-cds-rdm-events-stats-record-view-2024-01"
  }
}
DELETE cds-rdm-events-stats-record-view-2024-01

POST _aliases
{
  "actions": [
    {
      "remove": {
        "index": "deprecated-cds-rdm-dev-stats-record-view-2019-09",
        "alias": "cds-rdm-dev-stats-record-view"
      }
    }
  ]
}
```

Note, we have to follow this steps for all 4 indexes per year:
```
cds-rdm-events-stats-record-view-2025
[...]

PR #318, opened 2025-01-15
conf: update stats to yearly

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/256

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -21,7 +21,7 @@ from cds_rdm.permissions import (
 )
 from cds_rdm.files import storage_factory
 from celery.schedules import crontab
-from invenio_app_rdm.config import STATS_EVENTS as _APP_RDM_STATS_EVENTS
+from invenio_app_rdm.config import STATS_EVENTS as _APP_RDM_STATS_EVENTS, STATS_AGGREGATIONS as _APP_RDM_STATS_AGGREGATIONS
 from invenio_vocabularies.services.custom_fields import VocabularyCF
 from invenio_records_resources.services.custom_fields import KeywordCF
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -562,3 +572,92 @@ VOCABULARIES_NAMES_SCHEMES = {
 # We override the templates to add new fields needed for the migrated statistic events
 _APP_RDM_STATS_EVENTS["file-download"]["templates"] = "cds_rdm.stats.templates.events.file_download"
 _APP_RDM_STATS_EVENTS["record-view"]["templates"] = "cds_rdm.stats.templates.events.record_view"
+
+STATS_EVENTS = {
```
zzacharo, line 576: why do we need to dump the full config here?

invenio.cfg
```diff
@@ -562,3 +572,92 @@ VOCABULARIES_NAMES_SCHEMES = {
 # We override the templates to add new fields needed for the migrated statistic events
 _APP_RDM_STATS_EVENTS["file-download"]["templates"] = "cds_rdm.stats.templates.events.file_download"
 _APP_RDM_STATS_EVENTS["record-view"]["templates"] = "cds_rdm.stats.templates.events.record_view"
```
zzacharo, line 574: Also, take into account that we modify inline the `STATS_EVENTS` from invenio-app-rdm

invenio.cfg
jrcastro2: Thanks for the comment! I can simply override/add the params in the same way as done with the template and like that, there is no need to override the full conf ....

Issue #312, opened 2025-01-10
Show report numbers (and other fields?) in search results


ntarocco:
We should show the report number in the search results, as many users at CERN use that for searching.

What other fields should we display? Department, experiments, and the other subjects?

PR #341, opened 2025-01-30
resultList: add extra information

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/312
* Requires https://github.com/inveniosoftware/invenio-app-rdm/pull/2971

image

image

### DIFF
--- a/assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
+++ b/assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
@@ -1,0 +1,51 @@
+// This file is part of CDS RDM
+// Copyright (C) 2025 CERN.
+//
+// CDS RDM is free software; you can redistribute it and/or modify it
+// under the terms of the MIT License; see LICENSE file for more details.
+
+import _get from "lodash/get";
+import PropTypes from "prop-types";
+import React from "react";
+import { Item } from "semantic-ui-react";
+
+export const CDSRecordsResultsListItemDescription = ({ result, descriptionStripped }) => {
+    const getMetadataField = (path) => _get(result, path, []);
+
+    const cdsReferenceId = getMetadataField("metadata.identifiers")
+        .find((id) => id.scheme === "cds_ref")?.identifier;
+
+    const experiments = getMetadataField("custom_fields.cern:experiments").map((exp) => exp.title.en);
+    const accelerators = getMetadataField("custom_fields.cern:accelerators").map((acc) => acc.title.en);
+
+    const hasMetaData = cdsReferenceId || accelerators.length > 0 || experiments.length > 0;
+    const referenceOrAccelerators = cdsReferenceId || accelerators.length > 0;
+    return (
+        <>
+            <Item.Description className="truncate-lines-2">{descriptionStripped}</Item.Description>
+
+            {hasMetaData && (
+                <Item.Meta className="pt-20">
+                    {cdsReferenceId && <span className="mr-5">{cdsReferenceId}</span>}
+                    {accelerators.length > 0 && (
+                        <>
+                            {cdsReferenceId && <span className="ml- mr-5">|</span>}
+                            <span className={cdsReferenceId ? "ml-5 mr-5" : "mr-5"}>Accelerators: {accelerators.join(", ")}</span>
+                        </>
+                    )}
+                    {experiments.length > 0 && (
+                        <>
+                            {referenceOrAccelerators && <span className="ml-5 mr-5">|</span>}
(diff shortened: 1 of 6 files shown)


kpsherva:
@ntarocco do you think we could have an icon instead of `Accelerators:` ? I think there is too much text in one place and it does not help with quickly scanning the content. 
I would move the report number on top of the entry - either on the left side or inside a label. I think it is better to have it separately from the text, since it is an identifier. It will also help with text crowding on the bottom

zzacharo:
> @ntarocco do you think we could have an icon instead of `Accelerators:` ? I think there is too much text in one place and it does not help with quickly scanning the content. I would move the report number on top of the entry - either on the left side or inside a label. I think it is better to have it separately from the text, since it is an identifier. It will also help with text crowding on the bottom

+1 on the report number move. I am not sure about the icons, maybe after the move of the report number text is better? Also. @jrcastro2 from the screenshot do we have so many values at once for accelerator and experiments? Or did you add them as an example?

ntarocco:
> @ntarocco do you think we could have an icon instead of `Accelerators:` ? I think there is too much text in one place and it does not help with quickly scanning the content.

It is gonna be hard to find a good icon. If you add experiment/accelerator icons, that you can understand without reading, why not.

> I would move the report number on top of the entry - either on the left side or inside a label. I think it is better to have it separately from the text, since it is an identifier. It will also help with text crowding on the bottom

You missed our chat on this, IMO, at least for the time being, it should not be much visible. That's the entire objective. This might change in the future.

jrcastro2:
> > @ntarocco do you think we could have an icon instead of `Accelerators:` ? I think there is too much text in one place and it does not help with quickly scanning the content. I would move the report number on top of the entry - either on the left side or inside a label. I think it is better to have it separately from the text, since it is an identifier. It will also help with text crowding on the bottom
> 
> +1 on the report number move. I am not sure about the icons, maybe after the move of the report number text is better? Also. @jrcastro2 from the screenshot do we have so many values at once for accelerator and experiments? Or did you add them as an example?

Just added them as an example, I would guess most of them would not have so many, but since it accepts multiple value I tried it.

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
```diff
@@ -0,0 +1,51 @@
...
+    const cdsReferenceId = getMetadataField("metadata.identifiers")
+        .find((id) => id.scheme === "cds_ref")?.identifier;
+
+    const experiments = getMetadataField("custom_fields.cern:experiments").map((exp) => exp.title.en);
+    const accelerators = getMetadataField("custom_fields.cern:accelerators").map((acc) => acc.title.en);
+
+    const hasMetaData = cdsReferenceId || accelerators.length > 0 || experiments.length > 0;
+    const referenceOrAccelerators = cdsReferenceId || accelerators.length > 0;
+    return (
+        <div>
```
kpsherva, line 24:
```suggestion
        </>
```
<div> can be a problem for semantic styles

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
```diff
@@ -0,0 +1,51 @@
...
+                    {cdsReferenceId && <span className="mr-5">{cdsReferenceId}</span>}
+                    {accelerators.length > 0 && (
+                        <>
+                            {cdsReferenceId && <span className="ml- mr-5">|</span>}
+                            <span className={cdsReferenceId ? "ml-5 mr-5" : "mr-5"}>Accelerators: {accelerators.join(", ")}</span>
+                        </>
+                    )}
+                    {experiments.length > 0 && (
+                        <>
+                            {referenceOrAccelerators && <span className="ml-5 mr-5">|</span>}
```
kpsherva, line 38: at the first look the code looks not formatted, could you make sure we have prettier configured properly for cds-rdm?

jrcastro2: Hmm it seems correctly configured, we have the `.eslintrc.yml` with prettier and the github workflow for the testds is running it correclty, I run it locally and no errors were thrown

jrcastro2: Right, thanks for noticing, we were missing the assets folder to pass it to the linter, fixed!

run-js-linter.sh
```diff
@@ -28,4 +28,4 @@ for arg in $@; do
 done
 
 printf "${GREEN}Run eslint${NC}\n"
-npx eslint --no-error-on-unmatched-pattern -c .eslintrc.yml site/**/*.js
+npx eslint --no-error-on-unmatched-pattern -c .eslintrc.yml site/**/*.js assets/**/*.js
```
kpsherva, line 31: thanks a lot!

Issue #322, opened 2025-01-21
INSPIRE to CDS-RDM harvester: implement reader component


anikachurilova:
RFC: https://codimd.web.cern.ch/nawOikxYRjWef5y3s0whuA?view

ntarocco:
Where will the RFC be archived?

anikachurilova:
> Where will the RFC be archived?

@ntarocco here is the link to the RFC in the CERNBox: https://cernbox.cern.ch/external/eos/project/d/digital-repositories/Services/CDS/CDS-RDM/Migration/INSPIRE%20harvester/INSPIRE%20to%20CDS-RDM%20harvester.md?app=CodiMD&fileId=newproject-d%2170570974&contextRouteName=files-spaces-generic&contextRouteParams.driveAliasAndItem=eos/project/d/digital-repositories/Services/CDS/CDS-RDM/Migration/INSPIRE%20harvester

PR #344, opened 2025-02-03
INSPIRE harvester: add reader component

anikachurilova:
* configure and register the corresponding invenio-job
* create the Reader component
* draft of the Transformer component
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/322

All the args in the UI:
Screenshot 2025-02-03 at 17 53 52

### DIFF
--- a/site/cds_rdm/inspire_harvester/jobs.py
+++ b/site/cds_rdm/inspire_harvester/jobs.py
@@ -1,0 +1,149 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2025 CERN.
+#
+# CDS-RDM is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""Jobs module."""
+
+from datetime import timezone
+
+from invenio_i18n import gettext as _
+from invenio_vocabularies.jobs import ProcessDataStreamJob
+from marshmallow import Schema, ValidationError, fields, validates_schema
+from marshmallow_utils.fields import TZDateTime
+
+
+class InspireArgsSchema(Schema):
+    """Schema of task input arguments."""
+
+    since = TZDateTime(
+        timezone=timezone.utc,
+        format="%Y-%m-%d",
+        allow_none=True,
+        metadata={
+            "description": _(
+                "YYYY-MM-DD format. "
+                "Leave field empty if it should continue since last successful run."
+            )
+        },
+    )
+
+    until = TZDateTime(
+        timezone=timezone.utc,
+        format="%Y-%m-%d",
+        allow_none=True,
+        metadata={
+            "description": _(
+                "YYYY-MM-DD format. "
+                "End date of the date range. If this field is provided, then Since field is mandatory. Start/End date is included."
+            )
+        },
+    )
+
+    on = TZDateTime(
+        timezone=timezone.utc,
+        format="%Y-%m-%d",
+        allow_none=True,
+        metadata={"description": _("YYYY-MM-DD format. Harvest by exact date.")},
+    )
+
+    inspire_id = fields.String(allow_none=True)
+
+    job_arg_schema = fields.String(
+        metadata={"type": "hidden"},
+        dump_default="InspireArgsSchema",
+        load_default="InspireArgsSchema",
+    )
+
+    @validates_schema
+    def validate_date_range(self, data, **kwargs):
+        """Ensure that since <= until."""
+        since = data.get("since")
+        until = data.get("until")
+
+        if since and until and since > until:
+            raise ValidationError(
+                _("The 'Since' date must be earlier than or equal to the 'Until' date.")
+            )
+
+    @validates_schema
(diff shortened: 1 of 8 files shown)


site/cds_rdm/inspire_harvester/jobs.py
```diff
@@ -0,0 +1,99 @@
...
+
+    until = TZDateTime(
+        timezone=timezone.utc,
+        format="iso",
+        metadata={
+            "description": _(
+                "YYYY-MM-DD format. "
+                "End date of the date range. If this field is provided, then Since field is mandatory."
+            )
+        },
```
ntarocco, line 40: Missing help text if these dates are included or excluded in the range

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -0,0 +1,112 @@
...
+import requests
+from datetime import datetime
+from invenio_vocabularies.datastreams.readers import BaseReader
+from invenio_vocabularies.datastreams.errors import ReaderError
+
+
+class InspireHTTPReader(BaseReader):
+    """INSPIRE HTTP Reader."""
+
+    def __init__(self, origin=None, mode="r", since=None, until=None, on=None, inspire_id=None, *args, **kwargs):
```
ntarocco, line 20: what is the `mode` param?

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -0,0 +1,112 @@
...
+                    raise ReaderError(f"No results found when querying INSPIRE. See URL: {url}.")
+
+                yield response.content
+            else:
+                raise ReaderError(f"Error occurred while getting JSON data from INSPIRE. Error message: {response.text}")
+
+            # Get the next page URL if available
+            url = data.get("links", {}).get("next")
+
+    def read(self, item=None, *args, **kwargs):
```
ntarocco, line 56: Can you extract the validation part to a method? This will be more readable.
If not already done, can you give the validation block to an LLM and see if they come up with something more compact and readable?

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -0,0 +1,112 @@
...
+            response = requests.get(url, headers=headers)
+            data = response.json()
+
+            if response.status_code == 200:
+                if data["hits"]["total"] == 0:  # TODO make it a warning or info when we have proper logging
+                    raise ReaderError(f"No results found when querying INSPIRE. See URL: {url}.")
+
+                yield response.content
+            else:
+                raise ReaderError(f"Error occurred while getting JSON data from INSPIRE. Error message: {response.text}")
```
ntarocco, line 51: can you add the URL in the msg when it fails?

ntarocco (commented):
👏🏼 
Will tests come in a subsequent PR?

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def __init__(self, origin=None, mode="r", since=None, until=None, on=None, inspire_id=None, *args, **kwargs):
+        """Constructor."""
+        self._since = since
+        self._until = until
+        self._on = on
+        self._inspire_id = inspire_id
+        super().__init__(origin, mode, *args, **kwargs)
+
+    def _validate_date_format(self, date):
```
kpsherva, line 28: isn't date validated by the args schema? Do we need an additional validator?

site/cds_rdm/inspire_harvester/reader.py
anikachurilova: This is the "read" mode for opening the files needed for the parent class `BaseReader`. Actually now that I am looking at it, I don't see the reason for inheriting from `BaseReader`. It just adds additional complexity that we are not benefiting from. I will remove it

site/cds_rdm/inspire_harvester/reader.py
anikachurilova: No it is actually not validating the input data, so we need to add the manual validation

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -0,0 +1,135 @@
...
+                )
+
+            # Get the next page URL if available
+            url = data.get("links", {}).get("next")
+
+    def read(self, item=None, *args, **kwargs):
+        """Builds a query depending on the input data."""
+        self._validate_input_args()
+
+        document_type = "thesis"
```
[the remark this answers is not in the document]
anikachurilova: I had a discussion about this with Nico and there is no use for us to run the harvester for a set or specific document type. It would make sense if we had to run different types with different frequency.. but as it was in the old harvester, we want to just run everything at once.

So for now, as there is only theses, we need to specify it in the URL, but with all the next doc types, we will just remove this query parameter and run the harvester for all

kpsherva (approved):

zzacharo (approved):

zubeydecivelek (approved):

Issue #329, opened 2025-01-21
INSPIRE to CDS-RDM harvester: implement writer component


anikachurilova:
- Create a draft and publish newly created/updated record
- Provide proper logging and notification system

PR #375, opened 2025-03-03
INSPIRE harvester: implement minimal transformer and a writer

anikachurilova:
* map only required fields for now in the transformer
* change field type in marshmallow schema to Date
* split responsibilities in the transformer
* add writer component and tests to it
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/329


site/cds_rdm/inspire_harvester/jobs.py
```diff
@@ -7,19 +7,17 @@
...
 from marshmallow import Schema, ValidationError, fields, validates_schema
-from marshmallow_utils.fields import TZDateTime
 
 
 class InspireArgsSchema(Schema):
     """Schema of task input arguments."""
 
-    since = TZDateTime(
-        timezone=timezone.utc,
+    since = fields.Date(
```
anikachurilova, line 20: We don't really need to support timezone because our fields are just dates

site/cds_rdm/inspire_harvester/jobs.py
```diff
@@ -116,6 +112,9 @@ def build_task_arguments(
         cls, job_obj, since=None, inspire_id=None, until=None, on=None, **kwargs
     ):
         """Build task arguments."""
+        if isinstance(since, datetime):
+            since = since.date().strftime("%Y-%m-%d")
```
anikachurilova, line 116: Needed because `since` is a datetime when coming from the last successful run: https://github.com/inveniosoftware/invenio-jobs/blob/master/invenio_jobs/jobs.py#L105

site/cds_rdm/inspire_harvester/jobs.py
kpsherva, line 20: this field should be inherited from parent class (there is a generic schema implementing `since` field, and it should be used.
are you sure timezone is not needed?

site/cds_rdm/inspire_harvester/jobs.py
```diff
@@ -77,19 +73,19 @@ def validate_exclusive_arguments(self, data, **kwargs):
...
                 _(
-                    "When providing INSPIRE record ID for the search, all other args ('On', 'Since' and "
+                    "When providing INSPIRE record ID for the search, all other args ('On' and "
                     "'Until') are ignored. Please specify only inspire_id value."
                 )
             )
 
         # if `on` is provided, no other fields should be present
-        if on and any([inspire_id, since, until]):
+        if on and any([inspire_id, until]):
```
kpsherva, line 85: for readability, could we rename this variable?
```suggestion
        if on_date and any([inspire_id, until]):
```

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -0,0 +1,248 @@
...
+
+        return rdm_title, rdm_additional_titles
+
+    def _transform_publication_date(self):
+        """Mapping of INSPIRE thesis_info.date to metadata.publication_date."""
+        thesis_date = self.inspire_metadata.get("thesis_info", {}).get(
+            "date"
+        ) or self.inspire_metadata.get("thesis_info", {}).get("defense_date")
+        try:
+            parsed_date = str(parse_edtf(thesis_date))
```
kpsherva, line 115: why double conversion (edtf and then str) is needed?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -0,0 +1,248 @@
...
+                if inspire_title.get("subtitle"):
+                    rdm_additional_titles.append(
+                        {
+                            "title": inspire_title.get("subtitle"),
+                            "type": {
+                                "id": "subtitle",
+                            },
+                        }
+                    )
+            except Exception as e:
```
kpsherva, line 101: what kind of exceptions we might get here?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -0,0 +1,248 @@
...
+            self.metadata_errors.append(
+                f"Error occurred while parsing imprint.date to EDTF level 0 format for publication_date. INSPIRE "
+                f"record id: {self.inspire_metadata.get('control_number')}. Date: {thesis_date}. Error: {e}."
+            )
+            return None
+
+    def _transform_document_type(self):
+        """Mapping of INSPIRE document type to resource type."""
+        document_type = self.inspire_metadata.get("document_type")[0]
+        if document_type == "activity report":
```
kpsherva, line 127: could you improve the readability of these statements?

site/cds_rdm/inspire_harvester/jobs.py
anikachurilova: I had to redefine the field to be able to change the description (the one in the parent class has "YYYY-MM-DD HH:mm" format). 
Because in INSPIRE we can not search by datetime and only by date, I don't see the reason to support timezone

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -0,0 +1,248 @@
...
+        """Mapping of INSPIRE authors to creators and contributors."""
+        creators = []
+        authors = self.inspire_metadata.get("authors")
+        try:
+            for author in authors:
+                creators.append(
+                    {
+                        "person_or_org": {
+                            "type": "personal",
+                            "family_name": author.get("full_name"),
```
kpsherva, line 162: is there an option of distinguishing family name and given names?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -0,0 +1,248 @@
...
+                        }
+                    }
+                )
+            return creators
+        except Exception as e:
```
kpsherva, line 167: what kind of exceptions are you expecting here?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -0,0 +1,248 @@
...
+        resource_type = self._transform_document_type()
+        rdm_metadata["resource_type"] = {"id": resource_type}
+
+        # creators
+        creators = self._transform_creators()
+        rdm_metadata["creators"] = creators
+
+        # alternate identifiers
+        identifiers = self._transform_alternate_identifiers()
+        rdm_metadata["identifiers"] = identifiers
```
kpsherva, line 210: to improve readability
```suggestion
        title, additional_titles = self._transform_titles()
        return {
             "title": title,
             "additional_titles": additional_titles,
             "publication_date" : self._transform_publication_date(),
             "resource_type": self._transform_document_type(),
             "creators": self._transform_creators(),
             "identifiers": self._transform_alternate_identifiers()
         }
      
```

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,228 @@
...
+from invenio_rdm_records.proxies import current_rdm_records_service
+from invenio_records_resources.services.errors import FileKeyNotFoundError
+from invenio_vocabularies.datastreams.errors import WriterError
+from invenio_vocabularies.datastreams.writers import BaseWriter
+
+
+class InspireWriter(BaseWriter):
+    """INSPIRE writer."""
+
+    def __init__(self, *args, **kwargs):
```
kpsherva, line 24: constructor not needed?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,228 @@
...
+                current_rdm_records_service.delete_draft(system_identity, draft["id"])
+                return False
+
+        current_rdm_records_service.publish(system_identity, draft["id"])
+        return True
+
+    def _get_existing_records(self, inspire_id):
+        """Find records that have already been harvested from INSPIRE."""
+        # for now checking only by inspire id
+        return current_rdm_records_service.search(
```
kpsherva, line 163: how many records can search return? are you sure we are getting all of them by using this query?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,228 @@
...
+            if existing_records.total > 1:
+                raise WriterError(
+                    f"More than 1 record found with INSPIRE id {inspire_id}. CDS records found: {', '.join(existing_records_ids)}"
+                )
+            elif existing_records.total == 1:
+                handle_existing_success = self._handle_existing_records(
+                    entry, existing_records_ids, existing_records_hits, inspire_id
+                )
+                if not handle_existing_success:
+                    continue
```
kpsherva, line 60: could you explain why do we continue here? just for me to understand the logic (we can have a call)

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,228 @@
...
+                # no existing record in CDS - create and publish a new one
+                record_creation_success = self._create_new_record(inspire_id, entry)
+                if not record_creation_success:
+                    continue
+        return stream_entries
+
+    def _handle_existing_records(
+        self, entry, existing_records_ids, existing_records_hits, inspire_id
+    ):
+        existing_rec_id = existing_records_ids[0]
```
kpsherva, line 71: you could pass the first result already as an argument, so there would be no need to handle it here

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,228 @@
...
+    ):
+        """For records with updated files coming from INSPIRE, create and publish a new version."""
+        new_version_draft = current_rdm_records_service.new_version(
+            system_identity, existing_rec_id
+        )
+        draft_id = new_version_draft.to_dict()["id"]
+        files_enabled = new_version_draft.to_dict()["files"]["enabled"]
+
+        if files_enabled:
+            current_rdm_records_service.import_files(system_identity, draft_id)
```
kpsherva, line 121: are you sure we always import files from the previous version?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,228 @@
...
+                        if attempt < max_retries:
+                            current_app.logger.error("Retrying in 1 minute...")
+                            time.sleep(60)
+                except Exception as e:
+                    current_app.logger.error(
+                        f"Retrieving file request failed on attempt {attempt}. Max retries: {max_retries}. Error {e}. URL: {url}. Filename: {file['key']}. INSPIRE record id: {inspire_id}."
+                    )
+
+                    try:
+                        if attempt < max_retries:
```
kpsherva, line 219: not sure I understand why we are retrying both here and in line 210, could you explain?

site/tests/inspire_harvester/test_writer.py
```diff
@@ -0,0 +1,717 @@
...
+def _cleanup_record(recid):
+    """Delete a record after each test."""
+    current_rdm_records.records_service.delete(system_identity, recid)
+
+
+@pytest.fixture()
+def transformed_record_1_file():
+    """Transformed via InspireJsonTransformer record with 1 file."""
+    return {
+        "metadata": {
```
kpsherva, line 28: I understand the reader is not covered with any tests, we don't know if pagination would work etc. Do you mind adding a test which is mocking the Inspire response to test the whole workflow with almost real response object?

site/cds_rdm/inspire_harvester/transform_entry.py
anikachurilova: We don't know for sure what data will be coming from INSPIRE. For example, could be that `inspire_title.get("title")`  returns `None`. To me it's better if we will know about this

site/cds_rdm/inspire_harvester/transform_entry.py
anikachurilova: It was needed because `parse_edtf` returns `class 'edtf.parser.parser_classes.Date'` and I needed a string

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -0,0 +1,248 @@
...
+                return None
+
+        return {
+            "entries": rdm_files_entries,
+        }
+
+    def transform_files(self):
+        """Transform INSPIRE documents and figures."""
+        transformed_files = self._transform_files()
+        if transformed_files == {}:
```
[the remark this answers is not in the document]
anikachurilova: This has to specifically check for `{}` as referring to line 222 when there are no files

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,228 @@
...
+                    entry, existing_records_ids, existing_records_hits, inspire_id
+                )
+                if not handle_existing_success:
+                    continue
+            else:
+                # no existing record in CDS - create and publish a new one
+                record_creation_success = self._create_new_record(inspire_id, entry)
+                if not record_creation_success:
+                    continue
+        return stream_entries
```
kpsherva, line 66: please keep it for a separate PR to rework together, as a general idea:

```suggestion
    def write_many(self, stream_entries, *args, **kwargs):
        """Creates or updates the record in CDS."""
        entries = [entry.entry for entry in stream_entries]

        for entry in entries:
            identifiers = entry["metadata"].get("identifiers", {})
            inspire_id = next(
                (
                    item["identifier"]
                    for item in identifiers
                    if item["scheme"] == "inspire"
                ),
                None,
            )

            existing_records = self._get_existing_records(inspire_id)
            existing_records_hits = existing_records.to_dict()["hits"]["hits"]
            existing_records_ids = [hit["id"] for hit in existing_records_hits]
            
            multiple_matches_found = existing_records.total > 1
            should_update = existing_records.total == 1
            should_create = existing_records.total == 0

            if multiple_matches_found:
                raise WriterError(
                    f"More than 1 record found with INSPIRE id {inspire_id}. CDS records found: {', '.join(existing_records_ids)}"
                )
            if should_update:
                record_id = existing_records_ids[0]
                # why in your implementation you need to pass inspire_id as an argument?
                # I see this is only for reporting reason but only when creating a record - this can be avoided by creating correct exceptions, and the interface of the method will become clearer.
                self.update_record(record_id, entry)
                # no need for the if, the loop will continue either way
            if should_create:
                # no existing record in CDS - create and publish a new one
                try:
                   self.create_record(entry)
                except (PossibleException1, PossibleException2) as e:
                   raise WriterError("Failed to create record {inspire_id}"}
             
        return stream_entries
```

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def _handle_existing_records(
+        self, entry, existing_rec_id, existing_records_hits, inspire_id
+    ):
```
jrcastro2, line 66: I would add a docstring, describing a bit the logic, it helps on understanding it :)

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -0,0 +1,223 @@
...
+                self._create_files(only_in_entry, draft_id, inspire_id)
+            except WriterError:
+                current_rdm_records_service.delete_draft(system_identity, draft_id)
+                return False
+
+        if only_in_existing:
+            for key in only_in_existing:
+                current_rdm_records_service.draft_files.delete_file(
+                    system_identity, draft_id, key
+                )
```
jrcastro2, line 134: Isntead of importing and creating the ones from the entry and deleting the ones we imported and are not there, can we just simply create the ones being harvested? (similar to how is done in _create_new_record)

site/cds_rdm/inspire_harvester/writer.py
jrcastro2: I have the same doubt 😅

PR #406, opened 2025-03-20
Fix jobs

kpsherva:


ntarocco (approved):

site/cds_rdm/jobs.py
jrcastro2, line 1: Minor: Would it make sense to fix this in the interna method [`_build_task_arguments`](https://github.com/inveniosoftware/invenio-jobs/blob/master/invenio_jobs/jobs.py#L106)?

kpsherva: yes definitely, but this is easier for now

PR #438, opened 2025-04-16
search: add mappings for custom fields

ntarocco:
Before:

Screenshot 2025-04-16 at 17 55 16

After:

Screenshot 2025-04-16 at 17 56 17

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -32,6 +32,10 @@ from invenio_rdm_records.config import (
     RDM_RECORDS_IDENTIFIERS_SCHEMES, always_valid, RDM_RECORDS_PERSONORG_SCHEMES,
     RDM_PERSISTENT_IDENTIFIERS, RDM_PARENT_PERSISTENT_IDENTIFIERS
 )
+from invenio_records_resources.services.records.queryparser import (
+    QueryParser,
+    SearchFieldTransformer,
+)
 from invenio_rdm_records.proxies import current_rdm_records_service as record_service
 from invenio_rdm_records.config import (
     RDM_PERSISTENT_IDENTIFIERS,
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -312,6 +316,24 @@ CDS_REDIRECTION_COLLECTIONS_MAPPING = {
     "CERN Students Projects": "",  # uuid
 }
 
+RDM_SEARCH = {
+    **deepcopy(RDM_SEARCH),
+    "query_parser_cls": QueryParser.factory(
+        mapping={
+            "experiment": "custom_fields.cern\:experiments.title.en",
```
ntarocco, line 323: Open question: when writing in the search box `experiment`, do you think that the user searches for the `id`, or for the `title`?

invenio.cfg
```diff
@@ -312,6 +316,24 @@ CDS_REDIRECTION_COLLECTIONS_MAPPING = {
     "CERN Students Projects": "",  # uuid
 }
 
+RDM_SEARCH = {
+    **deepcopy(RDM_SEARCH),
+    "query_parser_cls": QueryParser.factory(
+        mapping={
```
ntarocco, line 322: This is currently an issue, also happening in Zenodo [here](https://github.com/zenodo/zenodo-rdm/blob/master/invenio.cfg#L550).
We have recently introduced protected internal fields [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/c807b72a4e62054df3c90cbdbe36c9a305c962dc/invenio_rdm_records/config.py#L277), but when overriding the config variable, this is gone. This should be changed, and made internal, not overridable.

invenio.cfg
kpsherva: I would say title, more human friendly field

kpsherva (approved):

Issue #325, opened 2025-01-21
INSPIRE to CDS-RDM harvester: implement record transformation functionality


anikachurilova:
Implement all the transformation rules from [here](https://docs.google.com/spreadsheets/d/1YTQ-g2hDS80XFQsoLTVQxldz1TU9QCl-MiqoQotAGt8/edit?gid=0#gid=0) and form a new CDS-RDM record

PR #444, opened 2025-04-24
INSPIRE harvester: implement transformer rules

anikachurilova:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/325


invenio.cfg
```diff
@@ -344,6 +344,7 @@ RDM_FILES_DEFAULT_MAX_FILE_SIZE = 50 * 10 ** 9  # 50GB
 
 JOBS_ADMINISTRATION_ENABLED = True
 
+IDENTIFIERS_SCHEMES_TO_DROP = ["SPIRES", "HAL", "OSTI", "SLAC", "PROQUEST"]
```
kpsherva, line 347: this config should be inside the harvester, since it is harvester bussiness, I don't think we will change it without modifying the code so it should not be configuration variable

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -130,61 +170,257 @@ def _transform_publication_date(self):
...
         try:
             # add INSPIRE id
             identifiers.append({"identifier": str(inspire_id), "scheme": "inspire"})
+
+            # external_system_identifiers
+            external_sys_ids = self.inspire_metadata.get(
+                "external_system_identifiers", []
+            )
+            for external_sys_id in external_sys_ids:
+                schema = external_sys_id.get("schema").lower()
```
kpsherva, line 401: this one will fail with AttributeError if there is no schema

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -64,6 +68,8 @@ def build(self):
             "access": self._access(),
         }
 
+        if record.get("pids"):
+            rdm_record["pids"] = record["pids"]
```
kpsherva, line 72: are you sure we can assign the pids directly? do they have the same structre in inspire?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -113,11 +119,44 @@ def _transform_titles(self):
...
+        """Validate that record has only 1 imprint."""
+        imprints = self.inspire_metadata.get("imprints", [])
+        if imprints:
+            if len(imprints) > 1:
+                self.metadata_errors.append(
+                    f"More than 1 imprint found. INSPIRE record id: {self.inspire_metadata.get('control_number')}."
+                )
+                return False
+            return imprints[0]
+        return False
```
kpsherva, line 132: let's try to optimise for readability, how about avoiding nested conditional statements:
```suggestion
        if not imprints:
            return False
        if  len(imprints) > 1:
                self.metadata_errors.append(
                    f"More than 1 imprint found. INSPIRE record id: {self.inspire_metadata.get('control_number')}."
                )
                return False
        
        return imprints[0]
 
```

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -113,11 +119,44 @@ def _transform_titles(self):
...
+
+    def _transform_publisher(self):
+        """Mapping of publisher."""
+        imprint = self._validate_imprint()
+
+        if imprint:
+            return imprint.get("publisher")
+
+        return None
```
kpsherva, line 141:
```suggestion
        imprint = self._validate_imprint()

        return imprint.get("publisher")
```
the result is the same and we save some lines of code

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -127,12 +133,44 @@ def _transform_titles(self):
...
+
         thesis_date = self.inspire_metadata.get("thesis_info", {}).get(
-            "date"
-        ) or self.inspire_metadata.get("thesis_info", {}).get("defense_date")
+            "date",
+        ) or (
+            imprint.get("date")
+            if imprint
+            else self.inspire_metadata.get("thesis_info", {}).get("defense_date")
+        )
```
kpsherva, line 167: this condidtional statement is a bit hard to read. Can you simplify it?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +184,251 @@ def _transform_publication_date(self):
...
+        #     )
+        #     return None
+
+        # return document_type_mapping.get(document_type)
+
+        document_types = self.inspire_metadata.get("document_type", [])
+        for doc_type in document_types:
+            if doc_type == "article":
+                self.metadata_errors.append("Articles are not supported for now.")
+                return None
```
kpsherva, line 214: we should check in a demo what happens with multiple types document import

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +184,251 @@ def _transform_publication_date(self):
...
-        """Mapping of INSPIRE authors to creators and contributors."""
+        """Mapping of INSPIRE authors to creators."""
+        authors = self.inspire_metadata.get("authors", [])
         creators = []
-        authors = self.inspire_metadata.get("authors")
+        for author in authors:
+            inspire_roles = author.get("inspire_roles")
+            if not inspire_roles:
+                creators.append(author)
+            elif "author" in inspire_roles or "editor" in inspire_roles:
```
kpsherva, line 250: are we sure about editor being considered as creator of a record?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +184,251 @@ def _transform_publication_date(self):
...
+                if creator_affiliations:
+                    rdm_creatibutor["affiliations"] = creator_affiliations
+
+                if creator_identifiers:
+                    rdm_creatibutor["person_or_org"][
+                        "identifiers"
+                    ] = creator_identifiers
+
+                if role:
+                    rdm_creatibutor["role"] = role[0]
```
kpsherva, line 291: is it possible that one author has multiple roles? we could add them multiple times to the entry with different role each time, no?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +184,251 @@ def _transform_publication_date(self):
...
+
+        for author_id in author_ids:
+            author_schema = author_id.get("schema")
+            if author_schema in schemes_map.keys():
+                processed_identifiers.append(
+                    {
+                        "identifier": author_id.get("value"),
+                        "scheme": schemes_map[author_schema],
+                    }
+                )
```
kpsherva, line 319: just because I got it confused with jsonschema 😅 
```suggestion
            author_scheme = author_id.get("schema")
            if author_scheme in schemes_map.keys():
                processed_identifiers.append(
                    {
                        "identifier": author_id.get("value"),
                        "scheme": schemes_map[author_scheme],
                    }
                )
```

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +184,251 @@ def _transform_publication_date(self):
...
+
+        return processed_identifiers
+
+    def _transform_author_affiliations(self, author):
```
kpsherva, line 323: can you create a ticket to revisit this part when RORs are available on the record?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +184,251 @@ def _transform_publication_date(self):
...
+        """Transform copyrights."""
+        # format: "© {holder} {year}, {statement} {url}"
+        copyrights = self.inspire_metadata.get("copyright", [])
+        result_list = []
+
+        for cp in copyrights:
+            holder = cp.get("holder", "")
+            statement = cp.get("statement", "")
+            url = cp.get("url", "")
+            year = str(cp.get("year", ""))
```
kpsherva, line 345: how will the entry look like (rest API) if we have only statement and the rest of values empty? will it assign empty strings?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +184,251 @@ def _transform_publication_date(self):
...
             identifiers.append({"identifier": str(inspire_id), "scheme": "inspire"})
+
+            # external_system_identifiers
+            external_sys_ids = self.inspire_metadata.get(
+                "external_system_identifiers", []
+            )
+            for external_sys_id in external_sys_ids:
+                schema = external_sys_id.get("schema").lower()
+                if schema == "cds":
+                    schema = "lcds"
```
kpsherva, line 411: we will need to align with Inspire on how to treat new CDS pids, can you add a task to track this?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +184,251 @@ def _transform_publication_date(self):
...
+                value = external_sys_id.get("value")
+                if schema.upper() in IDENTIFIERS_SCHEMES_TO_DROP:
+                    continue
+                if schema not in RDM_RECORDS_IDENTIFIERS_SCHEMES.keys():
+                    self.metadata_errors.append(
+                        f"Unexpected schema found in external_system_identifiers. Schema: {schema}, value: {value}. INSPIRE record id: {inspire_id}."
+                    )
```
kpsherva, line 418: does this mean the record will not be imported at all?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -247,35 +515,101 @@ def transform_custom_fields(self):
             if x.get("legacy_name")
         ]
 
-        custom_fields["cern:accelerators"] = accelerators
-        custom_fields["cern:experiments"] = experiments
+        imprint = self._validate_imprint()
+
+        if accelerators:
+            custom_fields["cern:accelerators"] = accelerators
```
kpsherva, line 521: how do we map INSPIRE accelerators to CDS accelerators? do names match?

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -247,35 +515,101 @@ def transform_custom_fields(self):
...
+        if experiments:
+            custom_fields["cern:experiments"] = experiments
```
kpsherva, line 523: same for experiments

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -247,35 +515,101 @@ def transform_custom_fields(self):
...
+        isbns = self.inspire_metadata.get("isbns", [])
+        online_isbns = []
+        for isbn in isbns:
+            value = isbn.get("value")
+            valid_isbn = normalize_isbn(value)
+            if not valid_isbn:
+                self.metadata_errors.append(f"Invalid ISBN '{value}'.")
+            else:
+                if isbn.get("medium") == "online":
+                    online_isbns.append(valid_isbn)
```
kpsherva, line 538: this will be true for thesis, but not so much for books. Can you open a ticket to revisit this in the future?

site/tests/conftest.py
```diff
@@ -119,6 +120,13 @@ def app_config(app_config):
 
     app_config["JSONSCHEMAS_HOST"] = "localhost"
     app_config["BABEL_DEFAULT_LOCALE"] = "en"
+    app_config["IDENTIFIERS_SCHEMES_TO_DROP"] = [
```
kpsherva, line 123: please remove these from the config var (see the first comment)

site/cds_rdm/inspire_harvester/transform_entry.py
anikachurilova: Sure! I have also added a test for the case if a record has 2 doc types and one of them is article: `"document_type": ["thesis", "article"]`. It fails with an error in that case. See here: https://github.com/CERNDocumentServer/cds-rdm/pull/444/files#diff-fbdc1c27597a8be737b958bc042fdb7ea4b30e03a308f7b564b3705804b63f72R672

site/cds_rdm/inspire_harvester/transform_entry.py
anikachurilova: Yes, here is a formula from Micha (also confirmed with the team):
- no role -> creator
- author -> creator, role = author
- editor -> creator, role = editor
- supervisor -> contributor, role = supervisor

site/cds_rdm/inspire_harvester/transform_entry.py
anikachurilova: Discussed irl, decided to take the 1st item

site/cds_rdm/inspire_harvester/transform_entry.py
anikachurilova: It will look like this: "© All rights reserved" provided `statement = "All rights reserved"`

invenio.cfg
kpsherva: not adressed yet

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -127,12 +134,41 @@ def _transform_titles(self):
...
-            "date"
-        ) or self.inspire_metadata.get("thesis_info", {}).get("defense_date")
+        imprint = self._validate_imprint()
+
+        thesis_info = self.inspire_metadata.get("thesis_info", {})
+        thesis_date = (
+            thesis_info.get("date")
+            or (imprint.get("date") if imprint else None)
+            or thesis_info.get("defense_date")
+        )
```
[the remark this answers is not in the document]
jrcastro2: Added as dates, type other, description Defense date

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -127,12 +135,39 @@ def _transform_titles(self):
...
+        """Validate that record has only 1 imprint."""
+        imprints = self.inspire_metadata.get("imprints", [])
+
+        if not imprints:
+            return {}
+        if len(imprints) > 1:
+            self.metadata_errors.append(
+                f"More than 1 imprint found. INSPIRE record id: {self.inspire_metadata.get('control_number')}."
+            )
+            return {}
```
kpsherva, line 148: I think in the next PRs we should revise on what data structures we return and why. Returning an empty dict does not seem like a good practice

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -127,12 +135,39 @@ def _transform_titles(self):
...
+        """Mapping of publisher."""
+        imprint = self._validate_imprint()
+        return imprint.get("publisher")
+
     def _transform_publication_date(self):
         """Mapping of INSPIRE thesis_info.date to metadata.publication_date."""
-        thesis_date = self.inspire_metadata.get("thesis_info", {}).get(
-            "date"
-        ) or self.inspire_metadata.get("thesis_info", {}).get("defense_date")
+        imprint = self._validate_imprint()
```
kpsherva, line 159: this can be also addressed in a new PR. I think we can split validation from transformation, and not to run validation functions twice.

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +181,263 @@ def _transform_publication_date(self):
...
+
+        # return document_type_mapping.get(document_type)
+
+        document_types = self.inspire_metadata.get("document_type", [])
+        for doc_type in document_types:
+            if doc_type == "article":
+                self.metadata_errors.append("Articles are not supported for now.")
+                return None
+
+        # uncomment the part above for other doc types (thesis have only 1 possible doc type)
```
kpsherva, line 213: I think we should have a task for this

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +181,263 @@ def _transform_publication_date(self):
...
+            # external_system_identifiers
+            external_sys_ids = self.inspire_metadata.get(
+                "external_system_identifiers", []
+            )
+            for external_sys_id in external_sys_ids:
+                schema = external_sys_id.get("schema").lower()
+                if schema == "cds":
+                    schema = "lcds"
+                value = external_sys_id.get("value")
+                if schema.upper() in IDENTIFIERS_SCHEMES_TO_DROP:
```
kpsherva, line 422: this condition should also apply to persistent_identifiers field, in case there is one id with a scheme which we don't want to keep

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -218,72 +447,230 @@ def _transform_alternate_identifiers(self):
...
+        if inspire_accelerators:
+            mapped_accelerators = self._transform_accelerators(inspire_accelerators)
+            if mapped_accelerators:
+                custom_fields["cern:accelerators"] = mapped_accelerators
+        if inspire_experiments:
+            mapped_experiments = self._transform_experiments(inspire_experiments)
+            if mapped_experiments:
+                custom_fields["cern:experiments"] = mapped_experiments
+        if imprint and imprint.get("place"):
+            custom_fields["imprint:imprint"] = {"place": imprint.get("place")}
```
kpsherva, line 596: I think we could also improve the nested `if` for readability at some point (it does not need to be this PR)

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -218,72 +447,230 @@ def _transform_alternate_identifiers(self):
...
+
+        imprint_fields = custom_fields.get("imprint:imprint", {})
+
+        isbns = self.inspire_metadata.get("isbns", [])
+        online_isbns = []
+        for isbn in isbns:
+            value = isbn.get("value")
+            valid_isbn = normalize_isbn(value)
```
kpsherva, line 604: this is a repetition, we can just use the already normalized ISBN, no need to normalise twice

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -187,6 +189,19 @@ def _create_new_version(self, entry, record):
                 system_identity, new_version_draft.id
             )
 
+    def _add_community(self, draft):
+        """Add CERN Scientific Community to the draft."""
+        with db.session.begin_nested():
```
kpsherva, line 194: could you explain why we need begin_nested here and only here? I kinda have a feeling it should be used in a broader context

site/cds_rdm/inspire_harvester/transform_entry.py
jrcastro2: RIght, it is, but unless we cache it, I don't see how we can avoid it, as it's used in different places, for the record metadata and custom fields, which are the same level

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -218,72 +447,230 @@ def _transform_alternate_identifiers(self):
...
             for x in self.inspire_metadata.get("accelerator_experiments", [])
             if x.get("legacy_name")
         ]
 
-        custom_fields["cern:accelerators"] = accelerators
-        custom_fields["cern:experiments"] = experiments
+        thesis_info = self.inspire_metadata.get("thesis_info", {})
+        defense_date = thesis_info.get("defense_date")
+        if defense_date:
+            custom_fields["thesis:thesis"] = self._transform_thesis(thesis_info)
```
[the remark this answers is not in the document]
jrcastro2: Yes, there is some more, still needs to be confirmed where to be added, but is more still.

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -218,72 +447,230 @@ def _transform_alternate_identifiers(self):
 
     def _transform_abstracts(self):
         """Mapping of abstracts."""
-        abstract = self.inspire_metadata["abstracts"][0]["value"]
-        return abstract
+        abstracts = self.inspire_metadata.get("abstracts", [])
+        if abstracts:
+            return abstracts[0]["value"]
+        return None
```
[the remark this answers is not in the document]
jrcastro2: Right, for this case wit hthe abstracts we could prioritize based on the source if needed.

site/cds_rdm/inspire_harvester/transform_entry.py
jrcastro2: RIght, also not sure why not enabling it now already, in any case I have created the issue: https://github.com/CERNDocumentServer/cds-rdm/issues/498

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +181,263 @@ def _transform_publication_date(self):
...
-        if document_type not in document_type_mapping:
-            self.metadata_errors.append(
-                f"Error occurred while mapping document_type to resource_type. Couldn't fine a mapping "
-                f"rule for document_type {document_type}. INS"
-                f"PIRE record id: {self.inspire_metadata.get('control_number')}."
-            )
-            return None
+        for author in authors:
+            inspire_roles = author.get("inspire_roles", [])
+            if "supervisor" in inspire_roles:
```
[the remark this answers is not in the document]
jrcastro2: Right, here the issue: https://github.com/CERNDocumentServer/cds-rdm/issues/499

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +181,263 @@ def _transform_publication_date(self):
...
+        #     )
+        #     return None
+
+        # return document_type_mapping.get(document_type)
+
+        document_types = self.inspire_metadata.get("document_type", [])
+        for doc_type in document_types:
+            if doc_type == "article":
+                self.metadata_errors.append("Articles are not supported for now.")
+                return None
```
jrcastro2, line 211: change this to fail for all the rest that are not thesis

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -146,69 +181,263 @@ def _transform_publication_date(self):
...
+        ]
+        IDENTIFIERS_SCHEMES_TO_DROP = ["SPIRES", "HAL", "OSTI", "SLAC", "PROQUEST"]
+
         try:
+            # persistent_identifiers
+            persistent_ids = self.inspire_metadata.get("persistent_identifiers", [])
+            for persistent_id in persistent_ids:
+                schema = persistent_id.get("schema").lower()
+                value = persistent_id.get("value")
+                if schema not in RDM_RECORDS_IDENTIFIERS_SCHEMES.keys():
```
jrcastro2, line 403:
```suggestion
if schema.upper() in IDENTIFIERS_SCHEMES_TO_DROP:
   continue
```

site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -218,72 +447,230 @@ def _transform_alternate_identifiers(self):
...
         }
-        if additional_descriptions:
-            rdm_metadata.update({"additional_descriptions": additional_descriptions})
-        title, additional_titles = self._transform_titles()
-        rdm_metadata["title"] = title
-        if additional_titles:
-            rdm_metadata["additional_titles"] = additional_titles
 
-        return rdm_metadata
+        result = {k: v for k, v in rdm_metadata.items() if v}
```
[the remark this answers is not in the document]
jrcastro2: discussed IRL, aply the same for custom fields

PR #449, opened 2025-04-28
config: update OAISERVER_ID_PREFIX to repository.cern

### LINKED ISSUE #436: oai IDs have wrong domain name
the oai IDs are: oai:cds-rdm.com:kz4kp-h2j85. Shouldn't that be oai:repository.cern:kz4kp-h2j85?

yashlamba:
Close #436

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -277,6 +277,9 @@ USERPROFILES_EXTEND_SECURITY_FORMS = True  # automatically use user's email addr
 OAISERVER_ID_PREFIX = "127.0.0.1:5000"
 """The prefix that will be applied to the generated OAI-PMH ids."""
 
+OAISERVER_ADMIN_EMAILS = ["CHANGE_ME"]
+"""The email address of the OAI-PMH administrator."""
+
 # Invenio-Search
 # ==============
 SEARCH_INDEX_PREFIX = "cds-rdm-"
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -274,7 +274,7 @@ USERPROFILES_EXTEND_SECURITY_FORMS = True  # automatically use user's email addr
 
 # OAI-PMH
 # =======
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -274,7 +274,7 @@ USERPROFILES_EXTEND_SECURITY_FORMS = True  # automatically use user's email addr
 
 # OAI-PMH
 # =======
-OAISERVER_ID_PREFIX = "cds-rdm.com"
+OAISERVER_ID_PREFIX = "repository.cern"
```
kpsherva, line 277: thanks for fixing it!
ideally this configuration should point to localhost and be replaced by config vars on openshift environments.

yashlamba: Question: should this point to the "default" or the dev env config?

How do we look at the `invenio.cfg` that is committed? Is it the dev env config generally or the "default"?

zzacharo: if there is an env related variable then we commit the local or default in invenio.cfg and in our openshift templates we define the value for each environment respectively

kpsherva: it should point to localhost, like the other domain related variables, and in openshift we will overwrite with config
Could you change it to 127.0.0.1:5000 ? it will be ready to merge :)

invenio.cfg
```diff
@@ -274,9 +274,12 @@ USERPROFILES_EXTEND_SECURITY_FORMS = True  # automatically use user's email addr
...
 """The prefix that will be applied to the generated OAI-PMH ids."""
 
+OAISERVER_ADMIN_EMAILS = ["cds-team@cern.ch"]
```
kpsherva, line 280: thanks for finding this one too!
the default should be "CHANGE_ME" since we don't want to expose our emails in the GH repos if not necessary (worring about spam)

PR #459, opened 2025-05-07
components: add CDSResourcePublication to enforce scientific records …

zzacharo:
requires https://github.com/inveniosoftware/invenio-rdm-records/pull/2043
closes #431

PR #467, opened 2025-05-13
components: add component for system subject validation

### LINKED ISSUE #458: service component: prevent from edition or deletion of specific type of subject
This is in migration context. After we migrate thesis, we will have some records containing subjects like this: `collection:AIDA`. This is to mark records that in the past were indexed in two collections at the same time (legacy CDS).
Since we are still shaping the way we will organise content in the new CDS, we would like to preserve the information about the old collection somehow. The easiest seems to be having a specific type of subject (other ideas welcomed also!).
The problem is that the owner of the content should not be able to remove or modify it, it is kind of a system subject (admins should be able to modify or remove). I think the easiest way to achieve that would be a service component for draft edit action.

yashlamba:
Closes #458

### DIFF
--- a/site/cds_rdm/components.py
+++ b/site/cds_rdm/components.py
@@ -82,3 +82,46 @@ class CDSResourcePublication(ServiceComponent):
         """Submit draft metadata for review."""
         if is_thesis(record):
             self._validate_thesis_community(identity, record, errors=None)
+
+
+class SubjectsValidationComponent(ServiceComponent):
+    """Service component for subjects validation.
+
+    To be used with records which were formerly in multiple collections on CDS (legacy).
+    We tag those records with subject `collection:<collection name>` to be able to retrieve them in the future.
+    those subjects should not be modifiable by a regular user.
+    """
+
+    def _validate_subject_changes(self, identity, updated_data, original_data):
+        """Validate that the subject changes are allowed."""
+        if identity.user.has_role("administration"):
+            return
+        updated_collection_subjects = {
+            s["subject"] for s in updated_data if s["subject"].startswith("collection:")
+        }
+        original_collection_subjects = {
+            s["subject"]
+            for s in original_data
+            if s["subject"].startswith("collection:")
+        }
+        if updated_collection_subjects != original_collection_subjects:
+            raise ValidationError(
+                "Collection subjects cannot be updated.",
+                field_name="metadata.subjects",
+            )
+
+    def update_draft(self, identity, data=None, record=None, **kwargs):
+        """Validate subject changes on update."""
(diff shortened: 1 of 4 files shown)


site/cds_rdm/components.py
yashlamba, line 1: Where should this file be?

zzacharo: I have added it in the same location in https://github.com/CERNDocumentServer/cds-rdm/pull/459

site/cds_rdm/components.py
```diff
@@ -0,0 +1,49 @@
...
+from invenio_drafts_resources.services.records.components import ServiceComponent
+from marshmallow import ValidationError
+
+
+class SubjectsValidationComponent(ServiceComponent):
+    """Service component for subjects validation."""
+
+    def _validate_subject_changes(self, identity, updated_data, original_data):
+        """Validate that the subject changes are allowed."""
+        if identity.user.has_role("admin"):
```
ntarocco, line 19: This should be good enough for us, however in general a user could be attached to a role or an action, not always a role.
We normally check if the `identity` provides a Need, which is the expanded `superuser_access` action.

site/cds_rdm/components.py
```diff
@@ -0,0 +1,49 @@
...
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""CDS service component for subjects validation."""
+
+from invenio_drafts_resources.services.records.components import ServiceComponent
+from marshmallow import ValidationError
+
+
+class SubjectsValidationComponent(ServiceComponent):
+    """Service component for subjects validation."""
```
ntarocco, line 15: Can we add a more descriptive comment? In this way, we will remember why we have this and we are doing this. Ping @kpsherva

ntarocco (approved):

site/cds_rdm/components.py
```diff
@@ -0,0 +1,48 @@
...
+# CDS-RDM is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+"""CDS service components."""
+
+from invenio_drafts_resources.services.records.components import ServiceComponent
+from marshmallow import ValidationError
+
+
+class SubjectsValidationComponent(ServiceComponent):
+    """Service component for subjects validation."""
```
kpsherva, line 14:
```suggestion
    """Service component for subjects validation.
         
         To be used with records which were formerly in multiple collections on CDS (legacy).
         We tag those records with subject `collection:<collection name>` to be able to retrieve them in the future.
         those subjects should not be modifiable by a regular user.
    """
```

zzacharo (approved):

PR #483, opened 2025-05-16
clc: add clc sync message

jrcastro2:
image


site/cds_rdm/utils.py
```diff
@@ -284,3 +286,21 @@ def merge(self, name_source, name_dest):
...
+
+def get_clc_sync_entry(record):
+    """Get the CLC sync entry for the record.
+
+    :param record: The record to get the CLC sync entry for.
+    :return: The CLC sync entry.
+    """
+    try:
+        clc_sync_entry = current_clc_sync_service.read(
+            system_identity,
```
ntarocco, line 299: I would rather pass the current user here, and then change the `read` permission to allow anyone to read this. Otherwise, we are basically by-passing the read permissions.
WDYT?

invenio.cfg
```diff
@@ -509,3 +511,12 @@ _APP_RDM_STATS_EVENTS["record-view"]["params"]["suffix"] = "%Y"
...
 _APP_RDM_STATS_AGGREGATIONS["file-download-agg"]["params"]["index_interval"] = "year"
 _APP_RDM_STATS_AGGREGATIONS["record-view-agg"]["params"]["index_interval"] = "year"
+
+
+APP_RDM_RECORD_LANDING_PAGE_TEMPLATE = "cds_rdm/records/detail.html"
+
+# Replace the manage menu template
+APP_RDM_DETAIL_SIDE_BAR_TEMPLATES = [
+    "cds_rdm/records/manage_menu.html" if template == "invenio_app_rdm/records/details/side_bar/manage_menu.html" else template
+    for template in DEFAULT_SIDE_BAR_TEMPLATES
```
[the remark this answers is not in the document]
jrcastro2: Yes, because the idea is to replace the existing `manage_menu.html` template only. The goal is to avoid overriding the entire template which we could do, but I simply want to add new data attributes.

site/cds_rdm/clc_sync/services/permissions.py
```diff
@@ -38,7 +37,7 @@ class CLCSyncPermissionPolicy(BasePermissionPolicy):
     """Permission policy for CLCSync."""
 
     can_create = [CLCExporter(), Administration(), SystemProcess()]
-    can_read = [CLCExporter(), Administration(), SystemProcess()]
+    can_read = [CLCExporter(), Administration(), SystemProcess(), AnyUser()]
```
kpsherva, line 40: why does Any User need to read?

site/cds_rdm/templates/semantic-ui/cds_rdm/records/detail.html
```diff
@@ -0,0 +1,32 @@
+{#
+    Copyright (C) 2025 CERN.
+  
+    CDS-RDM is free software; you can redistribute it and/or modify it
+    under the terms of the MIT License; see LICENSE file for more details.
+#}
+
+{%- extends "invenio_app_rdm/records/detail.html" %}
```
kpsherva, line 8: this is the reference to my first comment about using include instead of extend, but we could discuss in person about how to shape it

site/cds_rdm/ext.py
```diff
@@ -32,6 +33,7 @@ def init_app(self, app):
         """Flask application initialization."""
         self.init_services(app)
         self.init_resources(app)
+        app.jinja_env.globals["get_clc_sync_entry"] = get_clc_sync_entry
```
kpsherva, line 36: in this way you are adding it to all the templates in the whole application, do we really want this ? as well as I think we are accessing the db for all the records, does it make sense to check only for specific resource types? otherwise you always need to check if record has sync entry, even though it will not be the case for the majority of them

ntarocco: > do we really want this?

Yes, unless you have better ideas. This is an easy way to get some date from the backend, without changing the view and inject more context.

> does it make sense to check only for specific resource types? 

Yes why not, it can be done in the service so that we can skip the call if not matching the resource type. @kpsherva do we know the resource types? @jrcastro2 WDYT?

kpsherva: anything that can exist in a paper form, so all publications but if you want to specify even more:
```
publication-book
publication-bookchapter
publication-conferencepaper
publication-conferenceproceedings
publication-journal
publication-journalarticle
publication-preprint
publication-report
publication-technicalnote
publication-thesis
publication-dissertation
publication-workingpaper
```

site/cds_rdm/templates/semantic-ui/cds_rdm/records/detail.html
jrcastro2: I don't think we can do that as with the include we cannot override blocks inside the temaplate, form the [docs](https://jinja.palletsprojects.com/en/stable/templates/#include):

> The included template can extend another template and override blocks in that template. However, the current template cannot override any blocks that the included template outputs.

site/cds_rdm/templates/semantic-ui/cds_rdm/records/manage_menu.html
```diff
@@ -0,0 +1,14 @@
+{#
+  Copyright (C) 2022-2024 CERN.
+  
+  Invenio RDM Records is free software; you can redistribute it and/or modify
+  it under the terms of the MIT License; see LICENSE file for more details.
+#}
+
+{% extends "invenio_app_rdm/records/details/side_bar/manage_menu.html" %}
```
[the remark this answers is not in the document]
jrcastro2: Same as above, we need to override a block.

site/cds_rdm/ext.py
jrcastro2: I adde da check for any publication. Let me know if that is good enough or if we should add specifically the ones you have shared here.

site/cds_rdm/templates/semantic-ui/cds_rdm/records/detail.html
jrcastro2: DIscussed IRL, this we keep

site/cds_rdm/templates/semantic-ui/cds_rdm/records/manage_menu.html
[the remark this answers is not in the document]
jrcastro2: Discussed IRL, removed in favor of using include in the parent template and simply pass it via config

PR #485, opened 2025-05-20
thesis: validate the student programme participation

kpsherva:
* closes: https://github.com/CERNDocumentServer/cds-rdm/issues/474

### DIFF
--- a/site/cds_rdm/components.py
+++ b/site/cds_rdm/components.py
@@ -19,6 +19,7 @@ from invenio_rdm_records.services.errors import ValidationErrorWithMessageAsList
 from invenio_records_resources.services.uow import TaskOp
 from marshmallow import ValidationError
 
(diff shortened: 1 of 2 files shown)


site/cds_rdm/components.py
```diff
@@ -50,6 +51,19 @@ def is_thesis(record):
...
 
+    def _validate_programme_participation(self, identity, record_or_draft, errors=None):
+        """Validate that user has selected a programme participation."""
+        if record_or_draft.get("custom_fields", {}).get("cern:programmes") is None:
+
+            error_message = _(
+                f"Select the student programme in which you have participated."
+            )
+            if errors is None:
+                errors = []
```
zzacharo, line 62: we should raise here no? Otherwise, if via REST API someone publishes directly without saving first then the validation will not kick in...

Issue #503, opened 2025-06-11
check if we can rename LHCB to LHCb in the experiments vocabulary


kpsherva:
request on behalf of LHCb secretariat

sakshamarora1:
Recipe:
```python
from invenio_vocabularies.proxies import current_service
from invenio_access.permissions import system_identity

term = current_service.read(id_=("experiments", "LHCB"), identity=system_identity)
if term:
    term.data["title"]["en"] = 'LHCb'

updated_term = current_service.update(id_=("experiments", "LHCB"), identity=system_identity, data=term.data)
```

PR #508, opened 2025-06-18
vocabularies: experiements: rename LHCB to LHCb

sakshamarora1:
fixes: #503


kpsherva (approved):
The changes need to be propagated to production in a correct way
Screenshot 2025-06-18 at 18 06 10
see the duplication on sandbox

sakshamarora1:
Seems like it was already duplicated in sandbox, I checked the DB and the Search in sandbox, there are 2 entries. We should delete one.

Issue #291, opened 2024-12-16
Export formats/BibTeX improvements


ntarocco:
*Feedback received by users*

> It is annoying to download an export format, then open it, and copy its content. It would be nice to also have the possibility to directly copy it, without download.

It requires to add a new endpoint or make [the existing one](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/views/records.py#L286) configurable so that the text can be returned instead of downloading a file.
Note: this button should be also added close to the DOI, so you can very quickly copy the DOI value.

> The BibTeX format should contain the all-versions DOI, and not the single version DOI. In general, we always want to cite the latest version.

This requires careful thoughts. First of all, BibTeX format does not support multiple DOIs. Then, for previous versions (not the latest), it might make sense that someone wants to cite that specific version.
An idea could be to provide multiple export format, for that specific version or for all versions. We also need to see if the DOI is included in other export formats.

PR #521, opened 2025-07-15
invenio.cfg: Add export option for bibtex all versions

sakshamarora1:
fixes: https://github.com/CERNDocumentServer/cds-rdm/issues/291


invenio.cfg
```diff
@@ -525,4 +526,15 @@ _APP_RDM_STATS_AGGREGATIONS["record-view-agg"]["params"]["index_interval"] = "ye
...
 APP_RDM_RECORD_LANDING_PAGE_TEMPLATE = "cds_rdm/records/detail.html"
 
-RDM_DETAIL_SIDE_BAR_MANAGE_ATTRIBUTES_EXTENSION_TEMPLATE = "cds_rdm/records/manage_menu.html"
\ No newline at end of file
+RDM_DETAIL_SIDE_BAR_MANAGE_ATTRIBUTES_EXTENSION_TEMPLATE = "cds_rdm/records/manage_menu.html"
+
+APP_RDM_RECORD_EXPORTERS = {
+    **RECORD_EXPORTERS,
+    "bibtex-allversions": {
+        "name": _("BibTeX (all versions)"),
```
ntarocco, line 534: Nice, however I don't think it is clear enough.

```suggestion
        "name": _("BibTeX (DOI for all versions)"),
```

On top of that, I would also override the default BibTeX item and rename it.

Something like this?
Screenshot 2025-07-16 at 11 28 39

Issue #552, opened 2025-08-21
Implement GitLab integration for CDS


palkerecsenyi:
`invenio-github` is being turned into a generic VCS provider in https://github.com/inveniosoftware/invenio-github/pull/190, along with new support for GitLab.

This will make it relatively easy to add GitLab support to CDS once that's merged, as we'll basically just have to make config changes (and also a little custom CDS-specific code).

PR #554, opened 2025-08-21
WIP: User ID validator for GitLab

palkerecsenyi:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/552

---

* The new OAuth remote for GitLab will use the same system as all other remotes (i.e. being registered via `invenio-oauthclient`). This will allow us to plug in custom handlers at various points. Notably, we can override the GitLab-specific [`account_info_serializer` function](https://github.com/inveniosoftware/invenio-github/pull/190/commits/1354e805268736a9c7a56568aa9eb44f1c572268#diff-e7111d63b6d13c1d6a2a3b3369b3afb91789b22ff924f58e930352255de95964R81-R92) and perform our own validation on the raw `userinfo` object, even before it gets turned into the generic object with significantly less information.
* GitLab returns an `identities` field in the `userinfo`, containing a list of external authentication providers the user has linked to their account, and the ID each provider uses to represent the user. 
  
    For example:

    ```json
    {
        "id": 1234,
        "username": "johnsmith",
        "identities": [
            {
                "provider": "openid_connect",
                "extern_uid": "johnsmith",
                "saml_provider_id": null,
            },
            {
                "provider": "ldapmain",
                "extern_uid": "cn=johnsmith,ou=users,ou=organic units,dc=cern,dc=ch",
                "saml_provider_id": null,
            },
            {
                "provider": "kerberos",
                "extern_uid": "johnsmith@CERN.CH",
                "saml_provider_id": null,
            },
        ],
        ...
    }
    ```
    
    In this case, we can see the GitLab user `johnsmith` has a CERN username of `johnsmith`. It is possible for the GitLab username to be different to the CERN username, so we cannot simply rely on the GitLab username.

* We already store the user's CERN username in the database as part of the `extra_data` field for the CERN/keycloak [`RemoteAccount`](https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/models.py#L29). We just have to look this up (based on the OAuth client ID of the CERN remote app) and compare the two values.

* This way, we ensure the user has logged in with the GitLab account associated with the same CERN account as their CDS account, which avoids some security issues and bugs.

### DIFF
--- a/site/cds_rdm/vcs/handlers.py
+++ b/site/cds_rdm/vcs/handlers.py
@@ -1,0 +1,45 @@
+from __future__ import annotations
+
+from flask_login import current_user
+from invenio_cern_sync.sso import cern_remote_app_name
+from invenio_oauthclient import current_oauthclient
+
+from cds_rdm.errors import (
+    GitLabIdentityNotFoundError,
+    KeycloakGitLabMismatchError,
+    KeycloakIdentityNotFoundError,
+)
+
+
+def gitlab_account_info_serializer(original_serializer):
+    def inner(remote, resp, user_info, **kwargs):
+        cern_client_id = current_oauthclient.oauth.remote_apps.get(
+            cern_remote_app_name
+        ).consumer_key
+
+        user_keycloak_id: str | None = None
+        for remote_account in current_user.remote_accounts:
+            if remote_account.client_id == cern_client_id:
+                user_keycloak_id = remote_account.extra_data.get("keycloak_id")
+
+        if user_keycloak_id is None:
+            raise KeycloakIdentityNotFoundError(current_user.id)
+
+        gl_user_id = str(user_info["id"])
+        gl_identities = user_info["identities"]
+        gl_extern_uid: str | None = None
+        for identity in gl_identities:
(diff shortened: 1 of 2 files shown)


kmk142789 (approved):

kmk142789 (approved):

site/cds_rdm/vcs/handlers.py
```diff
@@ -0,0 +1,79 @@
...
+            prov = identity["provider"]
+
+            # On CERN GitLab, you have one GitLab account for each Keycloak account, with secondary accounts
+            # being separate GitLab accounts. You can sign in to one account with either openid_connect or
+            # kereberos, with the latter being used e.g. on CERN-provisioned computers.
+            # If a user has only ever signed in on a CERN device they might only have the kerberos method
+            # available, so we need to ensure we accept it.
+            if prov == "openid_connect":
+                gl_extern_uid = identity["extern_uid"]
+            elif prov == "kerberos":
```
zzacharo: note: I dont understand this case, can you login to your Gitlab account via kerberos? how would a GL identity exist if you havent logged in on the GL interface?

PR #561, opened 2025-08-27
schemes: added scheme for indico links

0einstein0:


site/cds_rdm/schemes.py
```diff
@@ -59,6 +59,15 @@ def is_inspire_author(val):
     return inspire_author_regexp.match(val)
 
 
+def is_indico(val):
+    """Test if argument is an Indico ID.
+
+    Warning: Indico IDs are just integers, with no structure, so this function will
+    say any integer is an Indico id
+    """
+    return str(val).isdigit()
```
zzacharo, line 68: Can you try to check what happens if you add another id that is supposed to be an integer? For example, the CDS scheme... does the url be generated correctly?

Issue #567, opened 2025-09-04
Inspire Harvester: Explore and implement a solution to map multiple document types as resource type


jrcastro2:
In Inspire a document can have multiple document types while on CDS we only allow a resource type. We need to find a solution for this use case.

kpsherva:
to discuss:
- [ ] first version goes through review? what about other versions?

PR #570, opened 2025-09-09
Add document types

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/567

### DIFF
--- a/site/cds_rdm/inspire_harvester/transform_entry.py
+++ b/site/cds_rdm/inspire_harvester/transform_entry.py
@@ -17,6 +17,18 @@ from invenio_records_resources.proxies import current_service_registry
 from opensearchpy import RequestError
 from sqlalchemy.orm.exc import NoResultFound
 
+# Mapping from INSPIRE document types to CDS-RDM resource types
+INSPIRE_DOCUMENT_TYPE_MAPPING = {
+    "article": "publication-article",
+    "book": "publication-book",
+    "report": "publication-report",
+    "proceedings": "publication-conferenceproceeding",
+    "book chapter": "publication-section",
+    "thesis": "publication-dissertation",
+    "note": "publication-technicalnote",
+    "conference paper": "publication-conferencepaper",
+}
+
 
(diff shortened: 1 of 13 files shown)


site/cds_rdm/inspire_harvester/transform_entry.py
```diff
@@ -517,34 +570,48 @@ def _transform_additional_descriptions(self):
...
             current_app.logger.warning(
-                f"No result found when searching for '{term}' in the vocabulary '{vocab_type}'. INSPIRE record id: {self.inspire_metadata.get('control_number')}. Error: {e}."
+                f"[inspire_id={self.inspire_metadata.get('control_number')}] No result found when searching for '{term}' in the vocabulary '{vocab_type}'. INSPIRE record id: {self.inspire_metadata.get('control_number')}. Error: {e}."
             )
 
     def _transform_accelerators(self, inspire_accelerators):
         """Map accelerators to CDS-RDM vocabulary."""
         mapped = []
         for accelerator in inspire_accelerators:
+            if accelerator.startswith("CERN-"):
```
kpsherva, line 607: there is another problem with mapping the accelarators and experiments, I saw in the logs previously the format:
`CERN-<ACCELERATOR>-<EXPERIMENT>`
which complicates things, was this adressed?

jrcastro2: Hmm, this is a bit more complicated to detect IMHO. Usually you are right, we do have `CERN-<ACCELERATOR>-<EXPERIMENT>` but also there are cases like https://inspirehep.net/api/literature/2892909 where we have:
```
"accelerator_experiments": [
{
"record": {
"$ref": "https://inspirehep.net/api/experiments/1275752"
},
"legacy_name": "CERN-NA-062"
}
],
```

where NA-062 is an experiment, I guess we want to check all the possible combinations in this case.

jrcastro2:
Other than Eslint failing, tests are passing locally

Issue #522, opened 2025-07-17
Thesis and dissertation - merge the resource types


kpsherva:
- [ ] add resource type metadata check to dissertation too
- [ ] update the thesis resource type to be mapped to datacite's dissertation one
- [ ] needs to be checked with the rdm community
- [ ] update datacite entries for the thesis (replace with dissertation type which exists in datacite)

PR #573, opened 2025-09-10
inspire_harvester: rename publication-thesis->publication-dissertation

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-rdm/issues/522


scripts/metadata_checks.py
```diff
@@ -36,7 +36,7 @@
                 "type": "comparison",
                 "left": {"type": "field", "path": "metadata.resource_type.id"},
                 "operator": "==",
-                "right": "publication-thesis",
```
kpsherva, line 39: can you note it down for deployment to change the value in the DB (for the existing check already)? Do you have deployment recipe?

PR #600, opened 2025-10-27
fix(config): add CDS identifier back to related ids

kpsherva:
* chore: formatting, remove unused imports

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -11,7 +11,6 @@ import os
 
 from copy import deepcopy
 from datetime import datetime, timedelta
-from invenio_i18n import lazy_gettext as _
 
 from cds_rdm import schemes
 from cds_rdm.custom_fields import CUSTOM_FIELDS, CUSTOM_FIELDS_UI, NAMESPACES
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -414,51 +413,63 @@ RDM_FILES_DEFAULT_MAX_FILE_SIZE = 50 * 10 ** 9  # 50GB
...
+            "validator": schemes.is_indico,
+            "datacite": "INDICO"
+        },
+        "hdl": {
+            "label": _("Handle"),
+            "validator": schemes.is_handle,
+            "datacite": "HANDLE",
+        },
+        "lcds": {"label": _("CDS"),
+                 "validator": schemes.is_legacy_cds,
```
palkerecsenyi: On a side note, could we simplify this function a bit?

https://github.com/CERNDocumentServer/cds-rdm/blob/13ddbd428f01495c950780d7ab16c9bb1a3921b0/site/cds_rdm/schemes.py#L91-L94

This is just equivalent to `val.isdigit()` right? Not sure if there is a performance difference but I think the regex is a bit less readable. It's not super important right now though

PR #608, opened 2025-10-30
Feat/add gltf previewer

ntarocco:
Closes https://github.com/inveniosoftware/invenio-previewer/issues/235

---

Preview:
Screenshot 2025-10-30 at 17 59 32

### DIFF
--- a/site/cds_rdm/assets/semantic-ui/js/cds_rdm/previewers/gltf-previewer.js
+++ b/site/cds_rdm/assets/semantic-ui/js/cds_rdm/previewers/gltf-previewer.js
@@ -1,1 +1,125 @@
-// TODO-EV
+import * as THREE from "three";
+import { OrbitControls } from "three/addons/controls/OrbitControls.js";
+import { GLTFLoader } from "three/addons/loaders/GLTFLoader.js";
+
+class ThreeScene {
+  constructor(config) {
+    this.config = config;
+    this.container = config.container;
+
+    this.scene = null;
+    this.camera = null;
+    this.renderer = null;
+    this.controls = null;
+  }
+
+  init() {
+    this.initScene();
+    this.initCamera();
+    this.initRenderer();
+    this.initControls();
+    this.addEventListeners();
+    this.loadModel();
+    this.startRendering();
+  }
+
+  initScene() {
+    this.scene = new THREE.Scene();
+    // Add a hemisphere light to make unlit models visible
+    this.scene.add(new THREE.HemisphereLight(0xe0e0e0, 0x707070, 1));
+  }
+
+  initCamera() {
+    const { camera_x, camera_y, camera_z } = this.config;
+    const width = window.innerWidth;
+    const height = window.innerHeight;
+
+    this.camera = new THREE.PerspectiveCamera(50, width / height, 0.1, 1000);
+    this.camera.position.set(camera_x, camera_y, camera_z);
+  }
+
+  initRenderer() {
+    this.renderer = new THREE.WebGLRenderer({ antialias: true, alpha: true });
+    this.renderer.setPixelRatio(window.devicePixelRatio || 1);
+    this.renderer.setClearColor(this.config.background, 1);
+    this.renderer.setSize(window.innerWidth, window.innerHeight);
+    this.container.appendChild(this.renderer.domElement);
+  }
+
+  initControls() {
+    this.controls = new OrbitControls(this.camera, this.renderer.domElement);
+    this.controls.zoomSpeed = 0.5;
+  }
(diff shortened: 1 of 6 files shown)


site/cds_rdm/previewer/gltf_previewer.py
```diff
@@ -11,7 +11,7 @@
 from invenio_previewer.proxies import current_previewer
 from invenio_previewer.utils import dotted_exts
 
-previewable_extensions = ["gltf"]
+previewable_extensions = ["glb", "gltf"]
```
ntarocco: We need to get all the possible extensions

site/cds_rdm/webpack.py
```diff
@@ -19,6 +19,10 @@
                 "gltf_previewer_js": "./js/cds_rdm/previewers/gltf-previewer.js",
                 "gltf_previewer_css": "./less/cds_rdm/previewers/gltf-previewer.less",
             },
+            dependencies={
+                "three": "^0.180.0",
+                "three-addons": "^1.2.0",
```
palkerecsenyi: Not sure if there's any alternative but this was archived in 2021 and last publish 7 years ago. https://github.com/marcofugaro/three-addons.

site/cds_rdm/assets/semantic-ui/js/cds_rdm/previewers/gltf-previewer.js
```diff
@@ -1 +1,115 @@
...
+        },
+        (error) => {
+          console.error("An error occurred loading the GLB file:", error);
+        }
+      );
+    }
+
+    startRendering() {
+      const renderLoop = () => {
+        requestAnimationFrame(renderLoop);
```
palkerecsenyi: Should we move this to the end of `renderLoop` to avoid running the next update before the current one accidentally? It probably won't make a huge difference if the render is fast but I've usually seen this call at the end rather than the start.

site/cds_rdm/assets/semantic-ui/js/cds_rdm/previewers/gltf-previewer.js
```diff
@@ -1 +1,115 @@
...
+        },
+        (xhr) => {
+          if (percentText && xhr.total) {
+            const percentComplete = Math.round((xhr.loaded / xhr.total) * 100);
+            percentText.textContent = `${percentComplete}`;
+            console.debug(`${percentComplete}% loaded`);
+          }
+        },
+        (error) => {
+          console.error("An error occurred loading the GLB file:", error);
```
palkerecsenyi: Will this show a visible error to the user? E.g. if the user is on an unstable network it's quite likely that the load would fail so we should make sure it's communicated clearly

site/cds_rdm/assets/semantic-ui/js/cds_rdm/previewers/gltf-previewer.js
```diff
@@ -1 +1,115 @@
...
+      );
+    }
+
+    startRendering() {
+      const renderLoop = () => {
+        requestAnimationFrame(renderLoop);
+        this.controls.update();
+        this.renderer.render(this.scene, this.camera);
+      };
+      renderLoop();
```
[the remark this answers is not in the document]
palkerecsenyi: e.g. see the "Rendering the scene" example: https://threejs.org/manual/#en/creating-a-scene

palkerecsenyi:
I've updated the code to fix the small nit picks above. I also added a hemisphere light source so that models without any built-in lighting are still visible against the black background. Let me know if you would like any other changes

site/cds_rdm/assets/semantic-ui/js/cds_rdm/previewers/gltf-previewer.js
```diff
@@ -1 +1,125 @@
...
+}
+
+document.addEventListener("DOMContentLoaded", () => {
+  // Configuration
+  const config = {
+    container: document.getElementById("gltf-container"),
+    background: 0x000000,
+    camera_x: 5,
+    camera_y: 5,
+    camera_z: 10,
```
kpsherva: @palkerecsenyi out of curiosity, do you know why not (0,0,0)?

palkerecsenyi:
@tpmccauley I've made some updates, let me know what you think! I haven't implemented automatic scaling/camera positioning as I couldn't find a simple solution for this, and I don't see a way to do it without making the integration a lot more complex. I hope that's okay! We can leave it as-is for now and see if we get frequent user reports of the model being incorrectly scaled or positioned.

zzacharo (approved):

PR #646, opened 2025-12-10
add(schemes): HAL identifier

kpsherva:


site/cds_rdm/schemes.py
```diff
@@ -117,10 +126,14 @@ def cds():
...
+
+
+def hal():
+    """Define scheme for CDS."""
     return {
-        "validator": is_indico,
-        "normalizer": lambda value: value,
-        "url_generator": lambda scheme, value: f"https://indico.cern.ch/event/{value}",
+        "validator": is_hal,
+        "normalizer": lambda value: value.lower(),
```
ntarocco: Can we add the `url_generator`? It is nice to have it clickable

PR #650, opened 2025-12-11
[blocked]change(config): move config variables to one single place

kpsherva:


sakshamarora1 (changes requested):
I think we should move the config.py file to invenio.cfg instead. To keep in sync with other RDM repos. WDYT? Unless there's a solid reason to to move to config.py
Also, for the commit message, it should be `refactor:`, see https://www.conventionalcommits.org/en/v1.0.0/#specification

Issue #631, opened 2025-11-27
4 - Show related records on landing page (based on related identifiers)


jrcastro2:
When a record includes CDS PIDs in `related_identifiers`, display these related records on the landing page. Similar behavior exists in Zenodo (example: https://zenodo.org/records/17563017).

Tasks:
	•	Parse `related_identifiers` for CDS PIDs.
	•	Fetch and display related records on landing page, similar to what Zenodo has done.

We should render the local pids.
Add a search bar to filter out results.

PR #653, opened 2025-12-14
feat: introduce linked records section on record detail pages

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/631
image

### DIFF
--- a/site/cds_rdm/assets/semantic-ui/js/cds_rdm/linked-records/api/config.js
+++ b/site/cds_rdm/assets/semantic-ui/js/cds_rdm/linked-records/api/config.js
@@ -1,0 +1,37 @@
+// -*- coding: utf-8 -*-
+//
+// Copyright (C) 2025 CERN.
+//
+// CDS-RDM is free software; you can redistribute it and/or modify it under
+// the terms of the MIT License; see LICENSE file for more details.
+
+export const apiConfig = (endpoint, baseQuery) => ({
+  axios: {
+    url: endpoint,
+    timeout: 5000,
+    headers: {
+      Accept: "application/vnd.inveniordm.v1+json",
+    },
+  },
+  interceptors: {
+    request: {
+      resolve: (config) => {
+        // Modify the params to combine base query with user search
+        if (config.params) {
+          const userQuery = config.params.queryString;
+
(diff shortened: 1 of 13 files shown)


site/cds_rdm/assets/semantic-ui/js/cds_rdm/linked-records/api/config.js
```diff
@@ -0,0 +1,37 @@
+// -*- coding: utf-8 -*-
+//
+// Copyright (C) 2025 CERN.
+//
+// CDS-RDM is free software; you can redistribute it and/or modify it under
+// the terms of the MIT License; see LICENSE file for more details.
+
+export const apiConfig = (endpoint, baseQuery) => ({
```
kpsherva: just for my understanding: do we need the custom function for this? isn't there one already somewhere?

site/cds_rdm/views.py
```diff
@@ -60,3 +60,46 @@ def inspire_link_render(record):
...
+    if not query_parts:
+        return None
+
+    # Combine all query parts with OR
+    combined_query = " OR ".join(query_parts)
+
+    # Exclude the current record and only show published records
+    final_query = f'({combined_query}) AND is_published:true NOT id:"{record_id}"'
+
+    return final_query
```
kpsherva: is there any way to find sibling references too? fe
I am on detail page of record A.
I find record B references record A.
there is also record C which references B but not A.
Can I include record C in this search too? Should I? tbd, ping @zzacharo @ntarocco

site/cds_rdm/views.py
```diff
@@ -60,3 +62,62 @@ def inspire_link_render(record):
             )
         )
     return ret
+
+
+def get_linked_records_search_query(record):
```
zzacharo: @jrcastro2 Can we add a quick test for this?

site/cds_rdm/assets/semantic-ui/js/cds_rdm/linked-records/api/config.js
zzacharo: I checked the `hiddenParams` but it seems to work only for adding additional filter parameters...

PR #662, opened 2025-12-18
fix(redirector): Redirect to download for non-previewable files

sakshamarora1:


site/cds_rdm/legacy/redirector.py
```diff
@@ -58,9 +58,18 @@ def legacy_files_redirect(legacy_id, filename):
         record = get_record_by_version(parent_pid.pid_value, version)
     except PermissionDeniedError:
         return abort(403)
+    filename_ext = filename.split(".")[-1]
     # Directly download files from redirected link to replicate the `allfiles-` behaviour from legacy
     if filename.startswith("allfiles-"):
         url_path = record["links"]["archive"]
+    # If the file is not previewable, redirect to the file download link instead
+    elif not current_app.config["IIIF_FORMATS"].get(filename_ext, None):
+        url_path = url_for(
```
zzacharo: we dont have this link in the record links?

sakshamarora1: We have the content link
```
/api/records/<record_id>/files/<filename>/content
```
We can re-use that, or is a non-api link prefered?

site/cds_rdm/legacy/redirector.py
```diff
@@ -56,11 +58,23 @@ def legacy_files_redirect(legacy_id, filename):
     version = query_params.pop("version", None)
     try:
         record = get_record_by_version(parent_pid.pid_value, version)
```
zzacharo: What happens here if `version` is not passed in the params?

zzacharo (approved):

site/cds_rdm/legacy/redirector.py
sakshamarora1: `if not version` takes care of that, this was already there :D

PR #710, opened 2026-02-24
Update deps

jrcastro2:


palkerecsenyi:
We're already doing something similar in #700 btw, not sure if this is the same thing?

PR #732, opened 2026-03-11
Multiple doc types

### LINKED ISSUE #567: Inspire Harvester: Explore and implement a solution to map multiple document types as resource type

kpsherva:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/567


site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -6,607 +6,275 @@
...
+            import traceback
+            traceback.print_exc()
         except ValidationError as e:
             error_message = f"Validation error while processing entry: {str(e)}."
-        # except Exception as e:
-            # raise e
-            # error_message = f"Unexpected error while processing entry: {str(e)}."
+            import traceback
+            traceback.print_exc()
+        except Exception as e:
```
TahaKhan998, line 67: I think the generic except Exception case is a bit risky here. For WriterError and ValidationError we actually build an error_message and add it to stream_entry.errors, so the failed entry is tracked properly. But here we only print the traceback and then continue. So if _route() fails because of some unexpected bug, nothing gets added to stream_entry.errors, op_type can stay unset, and the function still returns the entry. That feels like it could leave it in a half-failed state without the failure being tracked clearly. Should we either add an explicit error there too?

site/cds_rdm/inspire_harvester/transform/splitter.py
```diff
@@ -0,0 +1,94 @@
...
+                    inspire_id=self.inspire_id,
+                    cds_rdm_id=self.cds_id,
+                )
+                mappers = self.policy.build_for(resource_type)
+                assert_unique_ids(mappers)
+                patches = [
+                    m.apply(self.inspire_record, version_ctx, self.logger)
+                    for m in mappers
+                ]
+
```
TahaKhan998, line 87: I might be misunderstanding this, but is there a risk of generating duplicate versions for the same target resource type here? split() loops over every raw INSPIRE document_type, and if two different doc types map to the same CDS resource_type, it looks like we would append two versions with the same target type. Then later the writer would process that same resource type more than once. I’m also wondering if the unused _PREPRINT_DOC_TYPES constant was meant to group some of these doc types together instead.

site/cds_rdm/inspire_harvester/transform/mappers/article.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def map_value(self, src_record, ctx, logger):
+        """Mapping of abstracts."""
+        src_metadata = src_record.get("metadata", {})
+        abstracts = src_metadata.get("abstracts", [])
+        for abstract in abstracts:
+            source = abstract.get("source", "").lower()
+            if source and source not in ["arxiv", "cds"]:
+                return abstract["value"]
+            return abstracts[0]["value"]
```
TahaKhan998, line 82: i think this may return too early. Because the fallback return abstracts[0]["value"] is inside the loop, it looks like we only really inspect the first abstract. So if a preferred non-arxiv/non-cds abstract appears later in the list, we would never reach it. Was the fallback meant to be outside the loop?

site/cds_rdm/inspire_harvester/transform/mappers/preprint.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def map_value(self, src_record, ctx, logger):
+        """Mapping of abstracts."""
+        src_metadata = src_record.get("metadata", {})
+        abstracts = src_metadata.get("abstracts", [])
+        for abstract in abstracts:
+            source = abstract.get("source", "").lower()
+            if source and source in ["arxiv", "cds"]:
+                return abstract["value"]
+            return abstracts[0]["value"]
```
TahaKhan998, line 81: I think this may have the same early return issue as article.py.

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+            record_dict = record.to_dict()
+            existing_files = record_dict["files"]["entries"]
+        if should_import_files:
+            current_rdm_records_service.import_files(system_identity, draft.id)
+            logger.debug(
+                f"Imported files to {draft.id} from previous version: {record.id}"
+            )
+
+        new_files = incoming_record["files"].get("entries", {})
```
sakshamarora1, line 141: was it not `.files`? and it should always exist right? Otherwise we might wanna do get to not error out.

site/cds_rdm/inspire_harvester/transform/mappers/article.py
sakshamarora1: Good catch! CC @kpsherva

site/cds_rdm/inspire_harvester/transform/mappers/files.py
```diff
@@ -28,52 +32,52 @@ def map_value(self, src_record, ctx, logger):
...
-                    f"File key: {src_checksum}. "
-                    f"File_url: {url}. "
-                    f"INSPIRE record id: {ctx.inspire_id}. Error: {e}."
-                )
+            if self.filter(file):
+                logger.debug(f"Processing file: {file.get('filename', 'unknown')}")
+                filename = file["filename"]
+                source = file.get("source")
+                url = file["url"]
+                if "pdf" not in filename:
```
sakshamarora1, line 40: I think it would be better to use `.endswith(".pdf")` so that we don't miss edge cases like "Fulltext_pdf_paper"

site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
```diff
@@ -23,6 +23,10 @@ class DOIMapper(MapperBase):
 
     id = "pids"
 
+    def filter(self, doi):
+        """Filter doi based on given criteria."""
+        return True
```
sakshamarora1, line 28: Is it added for a future use case?

site/cds_rdm/inspire_harvester/transform/splitter.py
```diff
@@ -0,0 +1,94 @@
...
+from cds_rdm.inspire_harvester.logger import Logger, hlog
+from cds_rdm.inspire_harvester.transform.config import mapper_policy
+from cds_rdm.inspire_harvester.transform.context import MetadataSerializationContext
+from cds_rdm.inspire_harvester.transform.resource_types import (
+    INSPIRE_DOCUMENT_TYPE_MAPPING,
+)
+from cds_rdm.inspire_harvester.utils import assert_unique_ids, deep_merge_all
+
+# Doc types that belong to the arXiv/preprint stream
+_PREPRINT_DOC_TYPES = frozenset({"report", "note", "activity report"})
```
zubeydecivelek, line 14: Is this used somewhere? Is it needed?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -6,607 +6,275 @@
...
-                    break
+        elif match_result.found:
+            logger.info(f"Matching record found: CDS#{match_result.record_pid}")
+            self._update_record(stream_entry, record_pid=match_result.record_pid)
+            return "update"
 
-        return result
+        else:
+            self._create_record(stream_entry)
+            return "create"
```
zubeydecivelek, line 97: Is it possible to have a new entry (first time harvest) with multiple resource types?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -6,607 +6,275 @@
...
-        logger.debug(f"New version metadata updated: {new_version_draft.id}")
-        try:
-            logger.debug("Publishing new version draft")
+        new_version_entry = deepcopy(update_metadata)
 
-            current_rdm_records_service.publish(system_identity, new_version_draft.id)
+        if "pids" in new_version_entry:
+            del new_version_entry["pids"]["oai"]
+            if new_version_entry["pids"]["doi"]["provider"] != "external":
+                del new_version_entry["pids"]["doi"]
```
zubeydecivelek, line 220: Is it possible to have an entry without `oai` or `doi`?

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+    """Handles file I/O, diffing, uploading, and deletion for draft records."""
+
+    def __init__(self, retry_config: RetryConfig = None):
+        """Constructor."""
+        self.retry_config = retry_config or RetryConfig()
+
+    def compute_diff(self, existing_files, new_files) -> FileDiff:
+        """Return the set difference between existing and new file checksums."""
+        existing_checksums = [value["checksum"] for value in existing_files.values()]
+        new_checksums = [value["checksum"] for value in new_files.values()]
```
zubeydecivelek, line 49: Since arxiv files dont have a checksum [here](https://github.com/CERNDocumentServer/cds-rdm/pull/732/changes#diff-1e1bd9a4c7f79e9e6e3653a3d49772271cd43eccd8b6ee53ef1c9bdc1c84a11cR47) and we generate the checksum when file uploaded, if we receive the same arxiv file it'll have None in the new_checksums if I understand correctly. Could this cause the same file to be treated as changed?

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+            logger.info(f"{len(diff.to_delete)} files successfully deleted.")
+
+            logger.debug("Creating new files")
+            for key, file in new_files.items():
+                if file["checksum"] in diff.to_add:
+                    logger.debug(f"Processing new file: {key}")
+                    inspire_url = file.get("source_url")
+                    file_content = self.fetch(inspire_url, logger)
+                    self._upload_file(draft, file, file_content, logger)
+            logger.info(f"{len(new_files)} files successfully created.")
```
zubeydecivelek, line 161:
```suggestion
            logger.info(f"{len(diff.to_add)} files successfully created.")
```

site/cds_rdm/inspire_harvester/transform/mappers/contributors.py
```diff
@@ -38,16 +40,31 @@ def _transform_author_identifiers(self, author):
...
+                if ai.get("schema") == "ROR"
+            ]
+
+        for i, affiliation in enumerate(affiliations):
+            if i < len(ror_ids):
+                mapped_affiliations.append({"id": ror_ids[i]})
+            else:
+                value = affiliation.get("value")
+                if value:
+                    mapped_affiliations.append({"name": value.rstrip(".").strip()})
```
zubeydecivelek, line 67: Is it possible for `affiliations` and `ror_ids` to have different lengths? If so, could we end up assigning the wrong ROR to an affiliation because the mapping relies on the index?

site/cds_rdm/inspire_harvester/transform/mappers/preprint.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def filter(self, file):
+        """Filters files based on given criteria."""
+        material = file.get("material")
+        source = file.get("source")
+        if not (source or material):
+            return False
+        if (not material or material == "preprint") and source == "arxiv":
+            return True
+        if source == "CDS":
```
zubeydecivelek, line 31:
```suggestion
        if source.lower() == "cds":
```
isn't it safer like this?

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+        else:
+            should_import_files = (
+                record
+                and import_files
+                and record.data.get("files", {}).get("enabled", False)
+            )
+            record_dict = record.to_dict()
+            existing_files = record_dict["files"]["entries"]
+        if should_import_files:
+            current_rdm_records_service.import_files(system_identity, draft.id)
```
palkerecsenyi, line 136: would it not make sense to use `uow` on the service method calls? or is it not necessary here?

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+
+import requests
+from invenio_access.permissions import system_identity
+from invenio_rdm_records.proxies import current_rdm_records_service
+from invenio_records_resources.services.errors import FileKeyNotFoundError
+from invenio_vocabularies.datastreams.errors import WriterError
+
+
+@dataclass
+class RetryConfig:
```
palkerecsenyi, line 23: could we make this come from config variables? In case we need to alter it quickly after deploying

site/cds_rdm/inspire_harvester/transform/mappers/files.py
```diff
@@ -28,52 +32,52 @@ def map_value(self, src_record, ctx, logger):
...
-                    rdm_files_entries[filename]["metadata"] = file_metadata
-
-            except Exception as e:
-                ctx.errors.append(
-                    f"Error occurred while mapping files. "
-                    f"File key: {src_checksum}. "
-                    f"File_url: {url}. "
-                    f"INSPIRE record id: {ctx.inspire_id}. Error: {e}."
-                )
+            if self.filter(file):
```
palkerecsenyi, line 35: isn't this always true?

site/cds_rdm/inspire_harvester/update/engine.py
```diff
@@ -97,8 +98,10 @@ def update(self, current, incoming, ctx, logger):
 
         if conflicts or warnings:
             self.log_conflicts(conflicts, logger)
+
+        logger.debug(str(audit))
```
palkerecsenyi, line 102: is this left over?

site/cds_rdm/legacy/resolver.py
```diff
@@ -52,7 +52,10 @@ def get_record_by_version(parent_pid_value, version):
     if not version or version == "all" or latest_record["versions"]["index"] == version:
         return latest_record
     record_versions = get_record_versions(latest_record["id"])
-    if version not in record_versions.keys():
+    if str(version) not in record_versions.keys():
         # If record is not found, that means the version doesn't exist
         raise VersionNotFound(version=version, latest_record=latest_record)
-    return record_versions[version]
+    found_version = current_rdm_records_service.read(
```
palkerecsenyi, line 58: Why do we need this change?

site/tests/inspire_harvester/test_metadata_only_new_version.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2025 CERN.
```
palkerecsenyi, line 3: nit: some of the copyright headers are 2025 instead of 2026

invenio.cfg
```diff
@@ -112,6 +112,9 @@ TRUSTED_HOSTS = ['0.0.0.0', 'localhost', '127.0.0.1', 'localhost.cern.ch']
 
 # TODO: Set
 SQLALCHEMY_DATABASE_URI = "postgresql+psycopg2://cds-rdm:cds-rdm@localhost/cds-rdm"
+SQLALCHEMY_ENGINE_OPTIONS={
```
palkerecsenyi, line 115: We already have this set as an [environment variable](https://gitlab.cern.ch/cds-team/cds-rdm-openshift/-/blob/3a1418b2359b26ab64e2b94786213de6c634b659/configuration-dev.yaml#L68), where it has an additional config value.

site/cds_rdm/inspire_harvester/load/draft.py
```diff
@@ -0,0 +1,76 @@
...
+        """Update draft."""
+        return current_rdm_records_service.update_draft(
+            system_identity, draft.id, metadata
+        )
+
+    def new_version(self, record_pid):
+        """Create a new-version draft from an existing published record."""
+        return current_rdm_records_service.new_version(system_identity, record_pid)
+
+    def add_community(self, draft):
```
zzacharo, line 40: if we dont expect more type of communities to be added shall we rename to 
```suggestion
    def add_cern_research_community(self, draft):
```

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+    """Configuration for file fetch retries."""
+
+    max_retries: int = 3
+    retry_delay: int = 60  # seconds; only applied on network exceptions
+
+
+@dataclass
+class FileDiff:
```
zzacharo, line 31:
```suggestion
class FileChecksumsDiff:
```

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+            f" URL: {url}."
+        )
+        raise WriterError(
+            f"Failed to fetch file from {url} after {max_retries} retries."
+        )
+
+    def check_files_should_update(self, record, incoming_record, logger):
+        """Check if files should be updated."""
+        if not record:
+            return True
```
zzacharo, line 106: if there is not record, files should not be updated, I mean semantically no?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -6,607 +6,275 @@
...
         except WriterError as e:
             error_message = f"Error while processing entry : {str(e)}."
+            import traceback
+            traceback.print_exc()
         except ValidationError as e:
             error_message = f"Validation error while processing entry: {str(e)}."
-        # except Exception as e:
-            # raise e
-            # error_message = f"Unexpected error while processing entry: {str(e)}."
+            import traceback
```
zzacharo, line 65: repeatable import?

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def compute_diff(self, existing_files, new_files) -> FileDiff:
+        """Return the set difference between existing and new file checksums."""
+        existing_checksums = [value["checksum"] for value in existing_files.values()]
+        new_checksums = [value["checksum"] for value in new_files.values()]
+
+        return FileDiff(
+            to_add=list(set(new_checksums) - set(existing_checksums)),
+            to_delete=list(set(existing_checksums) - set(new_checksums)),
+            existing=list(set(existing_checksums)),
```
zzacharo, line 54: do we need the existing actually here?

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+            current_rdm_records_service.import_files(system_identity, draft.id)
+            logger.debug(
+                f"Imported files to {draft.id} from previous version: {record.id}"
+            )
+
+        new_files = incoming_record["files"].get("entries", {})
+        diff = self.compute_diff(existing_files, new_files)
+        should_update_files = bool(new_files) and bool(diff.to_add or diff.to_delete)
+        if should_update_files:
+            for filename, file_data in existing_files.items():
```
zzacharo, line 145: shall we move this functionality to the DraftLifecycle class as `delete_files`

zzacharo, line 145: potentially we could avoid the loop if the diff was returning more than the checksum e.g (filename, checksum)

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+            service.draft_files.init_files(
+                system_identity, draft.id, [file_data_to_init]
+            )
+            logger.debug(f"Filename: '{file_data['key']}' initialized successfully.")
+
+            service.draft_files.set_file_content(
+                system_identity, draft.id, file_data["key"], file_content
+            )
+            logger.debug(
+                f"Filename: '{file_data['key']}' content set successfully. Commit file..."
```
zzacharo, line 187:
```suggestion
                f"Filename: '{file_data['key']}' content set successfully. Committing file..."
```

site/cds_rdm/inspire_harvester/load/files.py
```diff
@@ -0,0 +1,210 @@
...
+
+        try:
+            if inspire_checksum is None:
+                # this can happen when we get the file directly from arxiv.
+                # unfortunately, arxiv does not expose checksums
+                del file_data["checksum"]
+            file_data_to_init = {
+                k: v for k, v in file_data.items() if k != "source_url"
+            }
+            service.draft_files.init_files(
```
zzacharo, line 178: For completeness: logger.debug(f"Filename: '{file_data['key']}' initializing.")

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@dataclass
+class MatchResult:
+    """Result of a record match attempt."""
+
+    ambiguous: bool = False
+    found: bool = False
+    record_pid: Optional[str] = None
+    matched_ids: List[str] = field(default_factory=list)
```
zzacharo, line 26: are these matched uuid or pids? If the latter,

```suggestion
    matched_record_pids: List[str] = field(default_factory=list)
```

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -0,0 +1,124 @@
...
+    def _build_filter_priority(self, entry, inspire_id, cdsrdm_id) -> OrderedDict:
+        """Build ordered filter dict for priority-based record lookup."""
+        doi = entry.get("pids", {}).get("doi", {}).get("identifier")
+        related_identifiers = entry["metadata"].get("related_identifiers", [])
+
+        cds_id = self._retrieve_identifier(related_identifiers, "cds")
+        arxiv_id = self._retrieve_identifier(related_identifiers, "arxiv")
+        report_number = self._retrieve_identifier(related_identifiers, "cdsrn")
+        return OrderedDict(
+            cds_pid={
```
zzacharo, line 78: Can we make this filters dataclasses with a meaningful name so it improves the readability? E.g

```
@dataclass(frozen=True)
class FilterCandidate:
    """A named search filter, tried in priority order to find a matching CDS record."""

    value: Optional[str]
    query: List[dsl.Q]

class ParentMatchFilter(FilterCandidate):
    """A named search filter, tried in priority order to find a matching CDS record."""

    value: Optional[str]
    query: List[dsl.Q]


return [
...,
ParentMatchFilter(value=cdsrdm_id,
                query=[dsl.Q("term", **{"parent.id": cdsrdm_id})],)
....
]
```

site/cds_rdm/inspire_harvester/load/files.py
kpsherva: these I've done on purpose in infinitive mode to be consistent everywhere with the tense used

site/cds_rdm/inspire_harvester/load/files.py
kpsherva: why? not sure I understand the suggestion

invenio.cfg
kpsherva: thats nice, thanks, but it was failing at that time in my local - if this is not needed in local anymore then we can remove

site/cds_rdm/inspire_harvester/writer.py
kpsherva: it is added on purpose - because the whole task fails if the error is not caught on one of the records - then no other record is processed, and it is a pity because we lose the whole update

site/cds_rdm/inspire_harvester/transform/mappers/article.py
kpsherva: this is exactly what we want - only the first abstract to be taken as main description, but I am happy you are paying a close attention to what is going on! thanks for the detailed review

site/cds_rdm/inspire_harvester/transform/mappers/preprint.py
kpsherva: same as above

site/cds_rdm/inspire_harvester/load/files.py
kpsherva: what do you mean about .files?
The entries don't always exist

site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
kpsherva: yes

site/cds_rdm/inspire_harvester/transform/splitter.py
kpsherva: could you add a test against it?

site/cds_rdm/inspire_harvester/writer.py
kpsherva: yes, but is there any suggestion behind this question?

site/cds_rdm/inspire_harvester/writer.py
kpsherva: yes, but is there any suggestion behind this question?

site/cds_rdm/inspire_harvester/load/files.py
kpsherva: yes we could create a test checking this - nice catch

site/cds_rdm/inspire_harvester/transform/mappers/contributors.py
kpsherva: not sure I fully understand what you mean but it would be good to have a test against this

site/cds_rdm/inspire_harvester/load/files.py
kpsherva: I am not sure how the errors would behave in this case - if we wrap the whole record creation and files upload in uow, then we fail the whole process in case of an error - so the record is not created at all, which potentially increases the amount of manual work which the curator needs to do later. If we separate the record creation from file upload, if the file upload fails, they only need to do the file correction, without inputting all the metadata.
Unless you meant uow for other reasons

site/cds_rdm/inspire_harvester/transform/mappers/files.py
kpsherva: for now yes, but it allows implementing a filter which will filter out some entries. this is done with other mappers if you want to check for examples

site/cds_rdm/inspire_harvester/update/engine.py
kpsherva: not a leftover, I want this to appear in debug level logs

site/cds_rdm/legacy/resolver.py
kpsherva: because the return type from the function changes based on the conditional branch. Once we return API object and other time ResultItem and it makes it harder to consume the info

site/cds_rdm/inspire_harvester/load/files.py
kpsherva: if there is no existing record then it means it is a new one so we want to upload files. I think the naming is unfortunate (record and incoming record - probably it should be existing record and incoming record)

site/cds_rdm/inspire_harvester/load/files.py
kpsherva: later on yes, if it happens that we need to alter it

site/cds_rdm/inspire_harvester/load/files.py
kpsherva: it gives the complete picture of the status of files in one object, is there any reason why not to include it?

site/cds_rdm/inspire_harvester/load/matcher.py
kpsherva: can be both external id, report number, PID, DOI - matching tries to find any identifier that exist

site/cds_rdm/inspire_harvester/writer.py
zubeydecivelek: I might be missing something here, but I `_create_record` reads ctx without using ctx["versions"], secondary resource types seem to be handled only in `_resource_type_versioning` on the update path. For a first-time harvest of a multi doc type record, would we only create the main entry?

site/cds_rdm/inspire_harvester/writer.py
zubeydecivelek: I was just asking if KeyError here acceptable?

site/cds_rdm/inspire_harvester/transform/mappers/contributors.py
zubeydecivelek: Mapping pairs affiliations with RORs by list index, but ror_ids is built by filtering only ROR entries out of affiliations_identifiers (skipping GRID etc.). So the index in ror_ids might not be the same as the index in affiliations_identifiers. I'm also not sure if it's a real case

Example where it could go wrong:
```python
"affiliations": [
  {"value": "University of Paris"},
  {"value": "CERN"}
],
"affiliations_identifiers": [
  {"value": "grid.xxx", "schema": "GRID"},
  {"value": "https://ror.org/01ggx4157", "schema": "ROR"}
]
```
After filtering, ror_ids = ["01ggx4157"] (only CERN's ROR).

The loop does:
i=0 → "University of Paris" → ror_ids[0] → CERN's ROR 
i=1 → "CERN" → i < len(ror_ids) is false → falls back to {"name": "CERN"}

PR #747, opened 2026-03-23
inspire: get exact match for accelerators and experiments

### LINKED ISSUE #551: [INSPIRE Harvester] Ensure vocabularies exact match
For any field that is mapped to a vocabulary in CDS, the value should be an exact match (instead of a search) and fail otherwise.

For example, [accelerators](https://github.com/CERNDocumentServer/cds-rdm/blob/master/site/cds_rdm/inspire_harvester/transform_entry.py#L538) and [experiments](https://github.com/CERNDocumentServer/cds-rdm/blob/master/site/cds_rdm/inspire_harvester/transform_entry.py#L552 should fail if the value is not found in the vocabulary, and not take the most similar first one. Ensure that this is applied to all vocabularies.

This means that we will probably need to have some mappings to transform INSPIRE experiments/accelerators/others to the CDS vocabularies' values. If possible, they could be a config so no code change is required when updating them.

Related to: https://github.com/CERNDocumentServer/cds-rdm/issues/377

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/551

Test from 15 May 2025

### DIFF
--- a/site/cds_rdm/config.py
+++ b/site/cds_rdm/config.py
@@ -55,3 +55,23 @@ CLC_URL_SYNC = "CHANGE_ME"
 
 CDS_ILS_IMPORTER_API_KEY = "CHANGE_ME"
 """API key for the CLC importer. This is a placeholder and should be replaced with a real key."""
+
+# =============================================================================
+# INSPIRE Harvester - Vocabulary Mappings
+# =============================================================================
+
+CDS_INSPIRE_ACCELERATOR_MAPPINGS = {
+    # Mappings from INSPIRE accelerator names to CDS vocabulary IDs
+    # Most INSPIRE accelerators match CDS exactly (e.g., "CERN LHC", "CERN SPS")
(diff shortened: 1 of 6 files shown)


kpsherva (commented):
I have suggested a simplification, please check. In addition, multiple tests are failing

site/cds_rdm/config.py
```diff
@@ -55,3 +55,23 @@
 
 CDS_ILS_IMPORTER_API_KEY = "CHANGE_ME"
 """API key for the CLC importer. This is a placeholder and should be replaced with a real key."""
+
+# =============================================================================
+# INSPIRE Harvester - Vocabulary Mappings
+# =============================================================================
+
+CDS_INSPIRE_ACCELERATOR_MAPPINGS = {
```
kpsherva: both variables are leftovers, should be removed

Issue #586, opened 2025-10-09
metadata-checks: require DOIs since the migration date (15th of May 2025)


kpsherva:
- [x] add to metadata checks to require DOIs for dissertation resource type
- [ ] mint DOIs for theses since the 15th od May to now (to be confirmed with SIS this is still desired)
- [ ] see https://github.com/CERNDocumentServer/cds-rdm/issues/426 for more context

kpsherva:
finally decided to drop the DOI requirement for thesis

TahaKhan998:
Recipe to revert the metadata check rule from the db:

```python
from copy import deepcopy
from invenio_checks.models import CheckConfig
from invenio_db import db

for check in CheckConfig.query.filter_by(check_id="metadata").all():
    params = deepcopy(check.params)
    params["rules"] = [
        rule for rule in params.get("rules", [])
        if rule.get("id") != "cern:dissertation/doi-required"
    ]
    check.params = params

db.session.commit()
```

PR #757, opened 2026-04-07
feat: require DOI for dissertations metadata check

TahaKhan998:
Part of https://github.com/CERNDocumentServer/cds-rdm/issues/586

This PR adds a metadata check to require a DOI for dissertation records.

A new rule was added to scripts/metadata_checks.py so that when the resource type is publication-dissertation, the metadata check validates that pids.doi.identifier is not empty.

## Recipe:

After web deployment — `metadata_checks`
  - Run the metadata-checks script to register the new metadata check that requires a DOI for dissertation records (`publication-dissertation`) on the CERN Scientific community:

    ```bash
    invenio shell scripts/metadata_checks.py
    ```

image

### DIFF
--- a/scripts/metadata_checks.py
+++ b/scripts/metadata_checks.py
@@ -52,6 +52,31 @@ CERN_RESEARCH_RULES = {
             ],
         },
         {
+            "id": "cern:dissertation/doi-required",
+            "title": "Dissertation DOI required",
+            "message": "Dissertations must provide a DOI",
+            "description": "To submit a dissertation, a DOI must be provided.",
+            "level": "error",
+            "error_path": "pids.doi",
+            "condition": {
+                "type": "comparison",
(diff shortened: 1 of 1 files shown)


zubeydecivelek (commented):
Looks good! 

Just a small thing: From your screenshot i dont see the error message
Image

Normally it looks like this :)
Image

TahaKhan998:
> Looks good!
> 
> Just a small thing: From your screenshot i dont see the error message Image
> 
> Normally it looks like this :) Image

Good catch, thanks. The validation was firing, but the field-level message wasn’t being anchored properly because the rule was missing error_path. I’ve added that now, so the DOI error shows up in the right place, and I also updated the screenshot.

palkerecsenyi (approved):

sakshamarora1 (approved):

PR #760, opened 2026-04-13
fix(harvester): align INSPIRE related identifier resource types

### LINKED ISSUE #727: bulk update related identifiers, resource type of INSPIRE ids in CERN research
The inspire id points to the record in inspire, so it should have the same resource type as the record itself (if the record is thesis, then the inspire ID in related_identifiers should be also thesis)

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/727
Related pr https://gitlab.cern.ch/cds-team/production_scripts/-/merge_requests/47
This PR updates INSPIRE related identifier mapping so metadata.related_identifiers[].resource_type.id matches the record’s detected resource type (ctx.resource_type.value) instead of always being hardcoded to publication-other.

Updated affected INSPIRE harvester tests and fixtures to reflect the new expected mapping behavior; updated test set passes.

### DIFF
--- a/site/tests/inspire_harvester/test_transformer.py
+++ b/site/tests/inspire_harvester/test_transformer.py
@@ -74,7 +74,7 @@ def test_transform_related_identifiers(mock_normalize_isbn, running_app):
         "identifier": "12345",
         "scheme": "inspire",
         "relation_type": {"id": "isvariantformof"},
-        "resource_type": {"id": "publication-other"},
+        "resource_type": {"id": "other"},
     } in result
     assert {
         "identifier": "978-0-123456-78-9",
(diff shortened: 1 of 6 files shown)


site/tests/inspire_harvester/test_transformer.py
```diff
@@ -74,7 +74,7 @@ def test_transform_related_identifiers(mock_normalize_isbn, running_app):
         "identifier": "12345",
         "scheme": "inspire",
         "relation_type": {"id": "isvariantformof"},
-        "resource_type": {"id": "publication-other"},
+        "resource_type": {"id": "other"},
```
sakshamarora1, line 77: Why is this just other now?

TahaKhan998: Because this test creates the context with resource_type=ResourceType.OTHER at [L59](https://github.com/CERNDocumentServer/cds-rdm/pull/760/changes/BASE..28b52f1d121eea7061ec089923472f67daffaba4#diff-fbdc1c27597a8be737b958bc042fdb7ea4b30e03a308f7b564b3705804b63f72R59). In site/cds_rdm/inspire_harvester/transform/resource_types.py, ResourceType.OTHER is explicitly defined as "other", so this assertion is validating the new intended behavior.

Image

kpsherva (approved):

PR #773, opened 2026-04-21
fix: improve harvester reports access and log downloads

### LINKED ISSUE #767: Sandbox: harvester notification email includes inaccessible "view full details" link
Sandbox harvester notification emails include a view full details link that recipients should not be able to access.

This link should be removed from the email template.

Image

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/767

## Email template, removed "View Full Details"

The email had a "View Full Details" button pointing to /administration/jobs/<job_id>/<run_id>, but that page requires administration-access, which most recipients don’t have → leads to a permission error.

Removed the button from both HTML and plain-text templates. Summary + safe links are still there.

## "View Changes" works for curators

Curators could access the Harvester Reports page, but "View Changes" failed due to can_search_revisions not including the harvester-curator path (blank page / 403).

Added HarvesterCurator to can_search_revisions, so curators can now actually see revision history.

## Download button now returns real run logs

Previously, download used audit logs - only showed what got published.

Now it queries JOBS_LOGGING_INDEX using run_id + job_id and formats output like the admin UI:

[timestamp] LEVEL message
grouped by task_id
includes run status + failure message
respects max results limit

The .log is now a proper offline copy of the run logs.

## Harvester Reports shows only system publishes

Page was showing all record.publish logs (including manual ones).

Now also filters user_id = "system" (backend + frontend), so it only shows harvester activity.

### DIFF
--- a/site/cds_rdm/administration/harvester_reports.py
+++ b/site/cds_rdm/administration/harvester_reports.py
@@ -125,5 +125,8 @@ class HarvesterReportsView(AdminResourceListView):
             headers=self.get_search_request_headers(**kwargs),
             pagination_options=(20, 50),
             default_size=20,
-            hidden_params=[["action", "record.publish"]],
+            hidden_params=[
+                ["action", "record.publish"],
+                ["user_id", "system"],
+            ],
         )
--- a/site/cds_rdm/harvester_download/resources/resource.py
+++ b/site/cds_rdm/harvester_download/resources/resource.py
@@ -7,13 +7,19 @@
 
 """Harvester download resource."""
 
+import re
+import uuid
 from datetime import datetime
 
-from flask import Response, g, request, stream_with_context
+from flask import Response, current_app, request
 from flask_resources import Resource, route
-from invenio_audit_logs.proxies import current_audit_logs_service
+from invenio_access.permissions import system_identity
+from invenio_jobs.models import Run
+from invenio_jobs.proxies import current_jobs_logs_service
 
 from cds_rdm.administration.permissions import curators_permission
+
+INSPIRE_HARVESTER_TASK = "process_inspire"
 
 
 class HarvesterDownloadResource(Resource):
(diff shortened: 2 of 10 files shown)


site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -27,44 +34,149 @@ def create_url_rules(self):
...
+        header.append("=" * 80)
+
+        logs = "\n".join(header + lines)
+
+        if not lines:
+            logs += "\n" + (run.message or "No logs available for this run.\n")
 
         timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
-        filename = f"harvester_logs_{timestamp}.txt"
+        filename = f"harvester_run_{run.id}_{timestamp}.log"
```
sakshamarora1:
```suggestion
        filename = f"harvester_logs_{run.id}_{timestamp}.log"
```

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -27,44 +34,149 @@ def create_url_rules(self):
...
+                return {"message": "Invalid harvester run query"}, 400
+            run_query = run_query.filter_by(finished_at=finished_at)
+        else:
+            run_query = run_query.filter_by(finished_at=None)
+
+        run = run_query.one_or_none()
+        if not run:
+            return {"message": "Run not found"}, 404
+
+        full_index_name = prefix_index(current_app.config["JOBS_LOGGING_INDEX"])
```
sakshamarora1: We should use the service layer here.

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -27,44 +34,149 @@ def create_url_rules(self):
...
+        status = getattr(run.status, "name", str(run.status))
+        header.append(f"Status: {status}")
+        header.append(f"Started: {_format_timestamp(run.started_at.isoformat())}")
+        if run.finished_at:
+            header.append(
+                f"Finished: {_format_timestamp(run.finished_at.isoformat())}"
+            )
+        if status in ("FAILED", "PARTIAL_SUCCESS"):
+            banner = "Job failed" if status == "FAILED" else "Job partially succeeded"
+            header.append("")
```
sakshamarora1, line 139: How does the end result look like? Can you attach an example file?

site/cds_rdm/permissions.py
```diff
@@ -71,6 +71,9 @@ class CDSRDMRecordPermissionPolicy(RDMRecordPermissionPolicy):
     can_create = [AuthenticatedRegularUser(), SystemProcess()]
     can_read = RDMRecordPermissionPolicy.can_read + [ArchiverRead()]
     can_search = RDMRecordPermissionPolicy.can_search + [ArchiverRead()]
+    can_search_revisions = RDMRecordPermissionPolicy.can_search_revisions + [
```
sakshamarora1, line 78: Why is this needed?

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: [harvester_logs_81eb2c75-b5c2-40a4-a3c1-7d4c375d6876_20260422_110359.log](https://github.com/user-attachments/files/26962284/harvester_logs_81eb2c75-b5c2-40a4-a3c1-7d4c375d6876_20260422_110359.log)
Check this

site/cds_rdm/permissions.py
TahaKhan998: So curators can use the "View Changes" button on the Harvester Reports page.

site/cds_rdm/administration/harvester_reports.py
```diff
@@ -125,5 +125,8 @@ def init_search_config(self, **kwargs):
             headers=self.get_search_request_headers(**kwargs),
             pagination_options=(20, 50),
             default_size=20,
-            hidden_params=[["action", "record.publish"]],
-        )
\ No newline at end of file
+            hidden_params=[
+                ["action", "record.publish"],
+                ["user_id", "system"],
```
kpsherva, line 130: did you check by any chance what does the REST API return for audit logs? we need to make sure REST API does not return wrong entries for curators role (on audit logs endpoint directly, not on this admin one)

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -27,44 +34,149 @@ def create_url_rules(self):
...
+        if not timestamp_match:
+            return {"message": "Invalid harvester run query"}, 400
+
+        start_time, end_time = timestamp_match.groups()
+        try:
+            started_at = datetime.fromisoformat(start_time)
+        except ValueError:
+            return {"message": "Invalid harvester run query"}, 400
+
+        run_query = Run.query.filter_by(started_at=started_at, parent_run_id=None)
```
kpsherva: I am not sure I understand the intention correctly - are you selecting the run based only on the started at time? if yes - I would make it more precise - it needs to be a run only of the specific harvester type of task (not a job - because a job is an instance of celery task, and if it is not created on the environment, then it will not exist)
otherwise, if there will be another run (of other type) which started the same time - unlikely but not impossible, then you will return logs of that other run

also, after additional thought... wouldn't it be better to rely on the run identifier rather than on it's timestamp? identifier would be more reliable, no?

site/cds_rdm/harvester_download/resources/resource.py
kpsherva: I agree

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -27,44 +34,149 @@ def create_url_rules(self):
...
+        # Admin UI header: run metadata + FAILED/PARTIAL_SUCCESS banner.
+        header = []
+        status = getattr(run.status, "name", str(run.status))
+        header.append(f"Status: {status}")
+        header.append(f"Started: {_format_timestamp(run.started_at.isoformat())}")
+        if run.finished_at:
+            header.append(
+                f"Finished: {_format_timestamp(run.finished_at.isoformat())}"
+            )
+        if status in ("FAILED", "PARTIAL_SUCCESS"):
```
kpsherva: I would include the success ones too, because it might contain warnings (or debug or info, depending on the settings), even if it was successful run

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -27,44 +34,149 @@ def create_url_rules(self):
...
+                return "N/A"
+            try:
+                return datetime.fromisoformat(
+                    raw.replace("Z", "+00:00")
+                ).strftime("%Y-%m-%d %H:%M")
+            except (ValueError, TypeError):
+                return raw
+
+        # Group by context.task_id in first-seen order (RunsLogs.js buildLogTree).
+        task_groups = {}
```
kpsherva, line 89: wasn't this grouping already done somewhere in the code? I am having flashbacks :)

site/cds_rdm/permissions.py
kpsherva: can you make sure they can only see revisions made by system user?
if they can see more, let's discuss on how to tackle it before jumping to implementations

site/cds_rdm/administration/harvester_reports.py
TahaKhan998: Yeah I checked the audit logs API. The issue was in HarvesterCurator.query_filter in generators.py. It was looking at identity.provides the wrong way so the filter never really worked and curators could call GET /api/audit-logs and get way too much back. I fixed it to use Permission(harvester_admin_access_action).allows(identity) and only return system user and record.publish like Harvester Reports. Updated test_harvester_curator_permissions too.

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: Yeah, the admin view already does that grouping in RunsLogs.js. The download is just plain text though, so I repeated the same task_id grouping here so the file looks like what you see in admin.

site/cds_rdm/permissions.py
kpsherva: discussed in person - we leave it as is - just for bookkeeping reasons, can you add a todo here to implement more restrictive permissions (plus please create an issue pointing to this todo, and cross-link)

TahaKhan998: done

PR #774, opened 2026-04-22
fix(redirector): Redirect to landing page if no request found

sakshamarora1:


site/cds_rdm/legacy/redirector.py
```diff
@@ -113,13 +113,25 @@ def legacy_comments_redirect(legacy_id):
     """Redirect legacy comments."""
     parent_pid = get_pid_by_legacy_recid(legacy_id)
     parent = RDMParent.pid.resolve(parent_pid.pid_value)
+    if parent.communities.default is None:
```
zzacharo: This cannot be as we require a community no?

zzacharo (approved):

PR #784, opened 2026-04-24
fix(redirector): improve error handling; switch to 302 redirects

palkerecsenyi:
- Created a new `FileNotFound` error for when a file cannot be found in any version of a record during the legacy file redirect.
- Created a template to communicate this to the user a bit more cleanly.
- Added tests for legacy redirects involving non-existent file names and for soft-deleted records.
- Added test for restricted record files during legacy redirect
- Changed legacy redirect HTTP status to [302](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/302) instead of [301](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/301) to clearly indicate to browsers that the target URL is subject to change (e.g. if a new record version is released) and should therefore not be put in a long-term cache.


site/tests/conftest.py
```diff
@@ -1395,7 +1395,7 @@ def minimal_restricted_record():
 
 @pytest.fixture()
 def minimal_record_with_files():
-    """Minimal record data as dict coming from the external world."""
+    """Minimal record data as dict coming from the external world, parametrized to test public and restricted files."""
```
sakshamarora1: leftover?

sakshamarora1 (approved):
LGTM!

site/cds_rdm/legacy/redirector.py
```diff
@@ -52,13 +52,25 @@ def version_not_found_error(error):
     )
 
 
+def file_not_found_error(error: FileNotFound):
```
zzacharo:
```suggestion
def legacy_file_not_found_error(error: FileNotFound):
```

site/cds_rdm/legacy/redirector.py
```diff
@@ -52,13 +52,25 @@ def version_not_found_error(error):
...
+    """Handler for the requested file not being found in any record version."""
+    return (
+        render_template(
+            "cds_rdm/file_not_found.html",
```
zzacharo:
```suggestion
            "cds_rdm/legacy_file_not_found.html",
```

site/cds_rdm/legacy/redirector.py
```diff
@@ -52,13 +52,25 @@ def version_not_found_error(error):
...
+
+
 def legacy_record_redirect(legacy_id):
     """Redirect legacy recid."""
     pid = get_pid_by_legacy_recid(legacy_id)
     url_path = invenio_url_for(
         "invenio_app_rdm_records.record_detail", pid_value=pid.pid_value
     )
-    return redirect(url_path, HTTP_MOVED_PERMANENTLY)
+    return redirect(url_path, HTTP_FOUND)
```
zzacharo: out of curiosity why the change of the response status?

palkerecsenyi: HTTP [301](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/301) (Moved Permanently) tells browsers to cache the redirect indefinitely, i.e. they will store the fact that the old URL is a "symlink" to the new one. So whenever the user visits the legacy URL after the first time, it will simply redirect them to the record page immediately without even sending a request to our server. Which is incorrect behaviour, because the record might have a new version released, and we always want to redirect them to the latest available version of the requested file/record.

HTTP [302](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/302) is normally not cached by most browsers and treated as a temporary redirect, which is more the behaviour that we want. It still isn't perfect but at least most browsers will behave correctly now.

We noticed this because the redirect was not working for @sakshamarora1 even after deploying the fix, and we found out that this redirect cache was the cause.

zzacharo: then what about also returning correct headers to instruct the browser not to cache the response?

palkerecsenyi: I've added the Cache-Control header :)

zzacharo (approved):

PR #791, opened 2026-04-28
add(vocabularies): video resource type

zubeydecivelek:
https://github.com/CERNDocumentServer/cds-migrator-kit/issues/386

### DIFF
--- a/app_data/vocabularies/resource_types.yaml
+++ b/app_data/vocabularies/resource_types.yaml
@@ -1089,3 +1089,21 @@
   tags:
     - depositable
     - linkable
+- id: video
+  icon: video
+  props:
+    csl: motion_picture
+    datacite_general: Audiovisual
+    datacite_type: ""
+    openaire_resourceType: "0033"
+    openaire_type: dataset
+    eurepo: info:eu-repo/semantics/other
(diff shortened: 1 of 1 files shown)


app_data/vocabularies/resource_types.yaml
```diff
@@ -1089,3 +1089,21 @@
   tags:
     - depositable
     - linkable
+- id: video
```
kpsherva, line 1092: now this is a tricky part... we can't just add it here, we need a way to allow selecting this only for related identifiers field, but not in the main resource type of the record itself

zubeydecivelek: it's already like this, [see](https://github.com/CERNDocumentServer/cds-rdm/pull/791/changes#diff-48344bf96d25537c97ce6d01ed2705096e0c7653be100cfe6c39fbd138094ca8R1108-R1109) it's only linkable not depositable

Screenshot 2026-05-13 at 10 20 57
Screenshot 2026-05-13 at 10 21 20

kpsherva: perfect, thanks!

PR #795, opened 2026-05-04
add(vocabularies): scientific committees, open access, former dep

### LINKED ISSUE #316: Open access values verification
in https://docs.google.com/spreadsheets/d/18tzNVo6sEUg-QZSj_EsBWkZC0stmV6daTgk_sTwxF0I/edit?gid=956730358#gid=956730358
under "Open access field" I placed a list of records which have 536__r filled with "Open access". Could you verify if these records need to be marked with open access?

I understood that open access is marked only when we somehow funded the article to provide open access, but if you see for example https://cds.cern.ch/record/198352 is this really the case? 
I am asking to understand how many of these values are correct and if having open access field is blocking for the migration of the IT collection.

### LINKED ISSUE #412: Propagate former departments on the migrated data
- [ ] add former divisions to the list of CERN departments
- [ ] update also IT collection (710)

kpsherva:
* needed for former experiments migration
* closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/412
* closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/316

Committees custom field will be needed for collections, it is not really a subject, since it is authority body. The committees will get their own communities too

## Departments
Screenshot 2026-05-02 at 18 33 13
Screenshot 2026-05-02 at 18 33 18
Screenshot 2026-05-02 at 20 00 02

## OA, section collapsed by default

Screenshot 2026-05-02 at 19 43 10
Screenshot 2026-05-02 at 19 43 18
Screenshot 2026-05-02 at 19 43 22


site/cds_rdm/app_data
```diff
@@ -0,0 +1 @@
+../../app_data
```
palkerecsenyi, line 1: Is this a symlink change? What is it doing?

site/cds_rdm/custom_fields/publishing.py
```diff
@@ -11,14 +11,60 @@
...
         # journal
-        *JOURNAL_CUSTOM_FIELDS_UI["fields"],
+        *JOURNAL_CUSTOM_FIELDS_UI["fields"] + [
+            dict(
+            field="cern:oa_level",
+            ui_widget="Dropdown",
+            props=dict(
+                label=_("Open Access Level"),
+                icon="lock open",
+                description=_("Select the open access level of this record."),
```
palkerecsenyi: I'm presuming users know what this refers to and what Bronze/Green/Gold open access means? Would it be worth adding a link to an explanatory website in the description, or is this something users already know about?

site/cds_rdm/custom_fields/cern.py
```diff
@@ -68,6 +74,21 @@
...
             ),
         ),
+        dict(
+            field="cern:committees",
+            ui_widget="Dropdown",
+            landing_page_search_attr="id",
+            props=dict(
+                label="Committee",
+                icon="users",
+                description="Please select a CERN committee related to this record if applicable.",
```
palkerecsenyi: nit: the descriptions of the custom fields all have slightly different formats, e.g. this one is "Please select .... if applicable", while `cern:administrative_unit` is "Optionally provide..." and `cern:programmes` is "Please select a ... applicable to your record". Maybe we could make them more consistent? It's not really a big problem though, just a nitpick :)

assets/templates/custom_fields/DepartmentDropdown.js
```diff
@@ -0,0 +1,134 @@
+// This file is part of Invenio-RDM-Records
+// Copyright (C) 2020-2025 CERN.
```
palkerecsenyi: nit:


```suggestion
// Copyright (C) 2026 CERN.
```

assets/templates/custom_fields/DepartmentDropdown.js
```diff
@@ -0,0 +1,134 @@
...
+//
+// Invenio-RDM-Records is free software; you can redistribute it and/or modify it
```
palkerecsenyi: nit:


```suggestion
// CDS-RDM is free software; you can redistribute it and/or modify it
```

assets/templates/custom_fields/CERNPublication.js
```diff
@@ -0,0 +1,66 @@
...
+      "custom_fields.cern:oa_level",
+      "custom_fields.cern:oa_funding_model",
+      "custom_fields.imprint:imprint",
+      "custom_fields.thesis:thesis",
+    ],
+  };
+
+  render() {
+    const { key, children, label, active } = this.props;
+    const [journal, oaLevel, oaFundingModel, imprint, thesis] = children;
```
palkerecsenyi, line 28: Why are we array-destructuring the children? This seems a little unusual

palkerecsenyi (commented):
Looks good!! Mostly just a question about the open access level field description

assets/templates/custom_fields/CERNPublication.js
kpsherva: what do you find as unusual?
The children are from the python config of custom fields, will always be sorted in the same way

site/cds_rdm/custom_fields/publishing.py
kpsherva: normally this will be filled only by the curators who know the values meaning, but it is a good point to make it more explicit that this field requires some expert knowledge (somehow). I will think about it

site/cds_rdm/app_data
kpsherva: commit has an explanation. I need to link the vocabularies folder to be packaged as part of the `./site` installation... because I don't want to use search for vocabularies values during migration. So I want the yamls to be accessible when I install cds-rdm in cds-migrator-kit, and this is the least disruptive way.
Not using the search saves us up to 1/3 to half a second (in some cases) per record, and when we are migrating 20k records then this becomes serious saving...

assets/templates/custom_fields/CERNPublication.js
palkerecsenyi: Ah so the `children` prop is not being used as the actual React `children` prop (e.g. a list of React nodes)? I was just thinking this is slightly unusual since `children` is sort of maybe a 'reserved' prop name. But if this is the name that we're already using elsewhere then it's fine

site/cds_rdm/custom_fields/publishing.py
kpsherva: @palkerecsenyi is it better now?

assets/templates/custom_fields/CERNPublication.js
kpsherva: yes, this is how the custom fields are propagating subfields in JS. It is not the best I agree with you but I didn't find a better way for now. We had this implementation in another local template of a CF (CERNFields)

zubeydecivelek (approved):
Looks good to me!

palkerecsenyi (approved):

zzacharo (approved):

PR #801, opened 2026-05-13
feat(admin): curator job run report and run log download

### LINKED ISSUE #789: feedback: harvester e-mail
- [x] bring back these links: 
Image

rename Download Harvester logs to "Download error log"
rename View Harvester reports to "View list of changes"

- [x] remove this button

Image

- [x] the downloadable report could be an html file resembling the job report in the admin panel (the goal is to keep the coloring because it helps reading it)

Image

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/789

- when an inspire harvester job finishes we email people. that email again has a “harvester actions” section with two links: one to grab a text log for that run (“download error log”) and one to open harvester reports filtered to that run (“view list of changes”)

- removed the old “view full details” button from that email — it pointed at the system jobs admin page and a lot of people on the mailing list don’t have access, so it was confusing

- new page: /administration/harvester-reports/<run_id>/report (curators only). shows the same kind of run summary + log lines as the admin job ui, with colours by level so it’s easier to read than raw text

- the existing download endpoint still returns a .log file; both that file and the html page use the real invenio jobs log search for the run, not a random audit log export

image

image

### DIFF
--- a/assets/less/cds-rdm/administration/harvester-reports.less
+++ b/assets/less/cds-rdm/administration/harvester-reports.less
@@ -54,3 +54,28 @@
     margin-top: 0.2em;
   }
 }
+
+.harvester-run-success-message {
+  white-space: normal;
+  text-align: left;
+}
+
+// Job run report: wrap long lines; scroll region height (matches admin log-table intent)
+.harvester-run-log-report {
+  width: 100%;
+
+  .harvester-run-log-table.log-table {
+    max-height: calc(100vh - 11rem);
+    overflow-x: hidden;
+  }
+
+  .harvester-run-log-segment {
+    overflow-x: hidden;
+  }
+
+  .log-line .log-message {
+    white-space: pre-wrap;
+    overflow-wrap: anywhere;
+    word-break: break-word;
+  }
+}
--- a/site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/DownloadButton.js
+++ b/site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/DownloadButton.js
@@ -5,39 +5,43 @@
 // under the terms of the GPL-2.0 License; see LICENSE file for more details.
 
 import React from "react";
-import { withState } from "react-searchkit";
 import { Button, Icon } from "semantic-ui-react";
 import { i18next } from "@translations/invenio_administration/i18next";
-import { extractRunIdFromQuery } from "./utils";
 
-const DownloadButtonComponent = ({ currentQueryState }) => {
-  const domContainer = document.getElementById("invenio-search-config");
-  const runs = JSON.parse(domContainer?.dataset.harvesterRuns || "[]");
-
-  const runId = extractRunIdFromQuery(
-    currentQueryState.queryString || "",
-    runs
-  );
-
-  const handleDownload = () => {
-    if (!runId) return;
-    const params = new URLSearchParams({ run_id: runId });
-    window.location.href = `/harvester-reports/download?${params.toString()}`;
-  };
-
-  return (
-    <Button
-      icon
-      labelPosition="left"
-      onClick={handleDownload}
-      disabled={!runId}
-      className="harvester-download-button"
-      size="small"
-    >
-      <Icon name="download" />
-      {i18next.t("Download")}
-    </Button>
-  );
-};
-
-export const DownloadButton = withState(DownloadButtonComponent);
+export const DownloadButton = ({ runId }) => (
+  <div className="ui horizontal list">
+    <div className="item">
+      <Button
+        icon
+        labelPosition="left"
+        onClick={() => {
+          if (!runId) return;
+          const q = new URLSearchParams({ run_id: runId });
+          window.location.href = `/harvester-reports/download?${q}`;
+        }}
+        disabled={!runId}
+        className="harvester-download-log-button"
+        size="small"
+      >
+        <Icon name="download" />
+        {i18next.t("Download error logs")}
+      </Button>
+    </div>
+    <div className="item">
+      <Button
+        icon
+        labelPosition="left"
+        onClick={() => {
+          if (!runId) return;
+          window.location.assign(`/administration/harvester-reports/${runId}/report`);
+        }}
+        disabled={!runId}
+        className="harvester-view-logs-button"
+        size="small"
+      >
+        <Icon name="file alternate outline" />
+        {i18next.t("View error logs")}
+      </Button>
+    </div>
+  </div>
+);
--- a/site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/SearchBar.js
+++ b/site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/SearchBar.js
@@ -28,18 +28,15 @@ const SearchBarComponent = ({ updateQueryState, currentQueryState }) => {
 
   const { sortOptions, sortOrderDisabled } = useContext(SearchConfigurationContext);
 
-  // Derive selected run from the timestamp in the current query — null if user typed a custom range
+  const [activeRunId, setActiveRunId] = React.useState(() => defaultRun?.id ?? null);
+
   const runIdFromQuery = extractRunIdFromQuery(currentQueryState.queryString, runs);
-  const selectedRun = runs.find((r) => r.id === runIdFromQuery) || null;
+  const effectiveRunId = runIdFromQuery || activeRunId;
+  const selectedRun = runs.find((r) => r.id === effectiveRunId) || null;
 
-  const [inputValue, setInputValue] = React.useState(currentQueryState.queryString || "");
-
-  // Auto-select default run on mount only if there is no existing query
-  React.useEffect(() => {
-    if (!currentQueryState.queryString && defaultRun) {
-      executeSearch(defaultRun, "");
-    }
-  }, []);
+  const [inputValue, setInputValue] = React.useState(
+    currentQueryState.queryString || ""
+  );
 
   const executeSearch = (run, userInput) => {
     const timestampFilter = buildTimestampFilter(run);
@@ -57,9 +54,23 @@ const SearchBarComponent = ({ updateQueryState, currentQueryState }) => {
     });
   };
 
+  React.useEffect(() => {
+    const q = currentQueryState.queryString || "";
+    const fromQuery = extractRunIdFromQuery(q, runs);
+    if (fromQuery) {
+      setActiveRunId(fromQuery);
+    } else if (!q && defaultRun) {
+      setActiveRunId(defaultRun.id);
+      executeSearch(defaultRun, "");
+    }
+  }, []);
+
   const onRunChange = (e, { value }) => {
+    setActiveRunId(value || null);
     const run = runs.find((r) => r.id === value);
-    executeSearch(run, "");
+    if (run) {
+      executeSearch(run, "");
+    }
   };
 
   const onBtnSearchClick = () => {
@@ -68,6 +79,10 @@ const SearchBarComponent = ({ updateQueryState, currentQueryState }) => {
       queryString: inputValue,
       hiddenParams,
     });
+    const id = extractRunIdFromQuery(inputValue, runs);
+    if (id) {
+      setActiveRunId(id);
+    }
   };
 
   const formatDate = (dateStr) => {
@@ -114,7 +129,7 @@ const SearchBarComponent = ({ updateQueryState, currentQueryState }) => {
             selection
             placeholder={i18next.t("Select a harvest run...")}
             options={runOptions}
-            value={selectedRun?.id || ""}
+            value={activeRunId || ""}
             onChange={onRunChange}
           />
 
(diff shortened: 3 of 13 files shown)


site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/DownloadButton.js
```diff
@@ -5,39 +5,39 @@
...
+        window.location.href = `/harvester-reports/download?${q}`;
+      }}
       disabled={!runId}
-      className="harvester-download-button"
+      className="harvester-download-log-button"
       size="small"
     >
       <Icon name="download" />
-      {i18next.t("Download")}
+      {i18next.t("Download run logs")}
```
kpsherva:
```suggestion
      {i18next.t("Download error logs")}
```
we need to be precise, otherwise I already saw some confusion on which logs report on failed records and which on the succeeded ones

site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/DownloadButton.js
```diff
@@ -5,39 +5,39 @@
...
+      onClick={() => {
+        if (!runId) return;
+        window.location.assign(`/administration/harvester-reports/${runId}/report`);
+      }}
+      disabled={!runId}
+      className="harvester-view-logs-button"
+      size="small"
+    >
+      <Icon name="file alternate outline" />
+      {i18next.t("View run logs")}
```
kpsherva:
```suggestion
      {i18next.t("View error logs")}
```

assets/less/cds-rdm/administration/harvester-reports.less
```diff
@@ -53,4 +53,58 @@
     color: #999;
     margin-top: 0.2em;
   }
-}
\ No newline at end of file
+}
+
+.harvester-report-actions {
```
kpsherva: in general, in css we try to avoid custom styling (with dedicated classes or assigned to html elements), and to rely on the semantic approach (class names related to behaviour and outcome, rather than specific dedicated element or class). For example in this case, the display flex class which you could assign to your element already exists in the theme and could be reused.

This approach helps us keeping the CSS maintenance to minimum and reduces the amount of stylesheets we write. Could you check how you can adapt your changes to our general approach?

site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/SearchBar.js
```diff
@@ -57,9 +54,25 @@ const SearchBarComponent = ({ updateQueryState, currentQueryState }) => {
...
+    const fromQuery = extractRunIdFromQuery(q, runs);
+    if (fromQuery) {
+      setActiveRunId(fromQuery);
+      return;
+    }
+    if (!q && defaultRun) {
+      setActiveRunId(defaultRun.id);
+      executeSearch(defaultRun, "");
+    }
+  }, []);
```
kpsherva, line 66:
```suggestion
  React.useEffect(() => {
    const q = currentQueryState.queryString || "";
    const runId = extractRunIdFromQuery(q, runs) || defaultRun.id;
    setActiveRunId(runId);
    }
    if (!q) {
      executeSearch(defaultRun, "");
    }
  }, []);
```
to improve readability. Why do we need to call execute search if q is not present?

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -22,6 +24,45 @@
...
 
+def _format_timestamp(value):
+    """``Run.started_at`` / ``finished_at``: naive DB datetimes are UTC → user locale."""
+    if value is None or value == "":
+        return "N/A"
+    if isinstance(value, datetime):
+        dt = (
+            value.replace(tzinfo=timezone.utc)
+            if value.tzinfo is None
+            else value
```
kpsherva: why is this part needed? we handle tz globally now if I remember correctly, no need to account for it in the code

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -22,6 +24,45 @@
...
+        )
+    else:
+        try:
+            dt = datetime.fromisoformat(str(value).replace("Z", "+00:00"))
```
kpsherva: same here

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -22,6 +24,45 @@
...
+        except (ValueError, TypeError):
+            return str(value)
+    return format_datetime(dt, "yyyy-MM-dd HH:mm", rebase=True)
```
kpsherva: what does rebase do? Normally we use other functions to serialize the time, could you check and adapt for consistency?

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -22,6 +24,45 @@
...
+
+def _format_log_hit_timestamp(value):
+    """Job log ``timestamp`` from search: zone-less ISO is local wall (like admin RunsLogs)."""
+    if value is None or value == "":
+        return "N/A"
+    if isinstance(value, datetime):
+        dt = value
+    else:
+        try:
+            dt = datetime.fromisoformat(str(value).replace("Z", "+00:00"))
```
kpsherva: same comment about timezone

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -22,6 +24,45 @@
...
+        )
+    else:
+        try:
+            dt = datetime.fromisoformat(str(value).replace("Z", "+00:00"))
+        except (ValueError, TypeError):
+            return str(value)
+    return format_datetime(dt, "yyyy-MM-dd HH:mm", rebase=True)
+
+
+def _format_log_hit_timestamp(value):
```
kpsherva: these two utility functions look very similar, why do we need both?

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -32,38 +73,33 @@ def create_url_rules(self):
...
-            return {"message": "Permission denied"}, 403
-
-        run_id = request.args.get("run_id", "").strip()
+    @staticmethod
+    def _resolve_harvester_run(run_id):
+        """Return ``(run, None)`` or ``(None, (error_dict, status))``."""
+        run_id = (run_id or "").strip()
         if not run_id:
-            return {"message": "Missing run_id"}, 400
+            return None, ({"message": "Missing run_id"}, 400)
```
kpsherva: there are specific flask or invenio adapted methods to serialize the REST responses in the correct way, could you check and use them?

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -147,9 +178,53 @@ def _format_timestamp(raw):
...
+            "truncation_message": truncation_message,
+            "lines": lines,
+            "error_count": error_count,
+        }
+        return ctx, None
+
+    def download(self):
+        """Download a harvester run's logs as a plain-text ``.log`` file."""
+        if not curators_permission.can():
+            return {"message": "Permission denied"}, 403
```
kpsherva: this response also could use a response serializer

site/cds_rdm/views.py
```diff
@@ -45,9 +45,28 @@ def create_cds_clc_sync_bp(app):
...
+    @report_bp.route("/administration/harvester-reports/<uuid:run_id>/report")
+    @curators_permission.require(http_exception=403)
+    def harvester_run_report(run_id):
+        ctx, err = ext.harvester_download_resource.report_template_context(str(run_id))
+        if err:
+            body, code = err
+            return jsonify(body), code
+        return render_template(
+            "cds_rdm/administration/harvester_run_report.html",
+            **ctx,
```
kpsherva: there is a bit of a mix of concepts here. In case of an error, you return a json, but in case everything is goes well, then we have a jinja template returned. Since this is the same view, an the UI side (not api) please use an error handler to return an error page - to keep the consistency of the returned formats

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: Makes sense, I’ll simplify this and remove the explicit timezone handling here so it follows the usual global formatting behavior.

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: Alright

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: Makes sense, I checked and rebase=True is the default already, so I’ll remove it here.
rebase just means “convert the datetime to the user’s timezone before formatting."

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: You’re right, after simplifying the timezone handling there isn’t really a meaningful difference anymore, so I’ll collapse them into a single formatter.

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: Done

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: I checked and switched this to use \HTTPJSONException for the REST errors instead of returning the raw dict/status tuples directly.

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: Yep, I switched this one to the same

assets/less/cds-rdm/administration/harvester-reports.less
TahaKhan998: Thanks, I updated this to rely on the existing Semantic UI/layout utility classes for the spacing and button alignment, and removed the dedicated action/spacing CSS. The remaining LESS is only for the log-specific behavior like wrapping long lines and controlling overflow.

site/cds_rdm/views.py
TahaKhan998: Makes sense, I switched this to return the normal themed HTML error page instead of JSON.

site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/SearchBar.js
TahaKhan998: I simplified the effect a bit, but I kept the initial executeSearch when there is no query because that is what actually applies the default run filter and triggers the first search. Without it we would only update the selected value in the dropdown, but the page would not load the initial filtered results.

site/cds_rdm/views.py
```diff
@@ -45,9 +45,27 @@ def create_cds_clc_sync_bp(app):
...
-    return ext.harvester_download_resource.as_blueprint()
+    bp = ext.harvester_download_resource.as_blueprint()
+
+    report_bp = Blueprint("cds_rdm_harvester_report_page", __name__)
+
+    @report_bp.route("/administration/harvester-reports/<uuid:run_id>/report")
+    @curators_permission.require(http_exception=403)
+    def harvester_run_report(run_id):
+        ctx, err = ext.harvester_download_resource.report_template_context(str(run_id))
+        if err:
```
kpsherva: to handle errors in more "flask" way please check .register_errror_handler of blueprint (there should be already examples in the code). Can it be applied here?

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -22,6 +24,20 @@
...
+        dt = value
+    else:
+        try:
+            dt = datetime.fromisoformat(str(value).replace("Z", "+00:00"))
+        except (ValueError, TypeError):
+            return str(value)
+    return format_datetime(dt, "yyyy-MM-dd HH:mm")
+
+
 class HarvesterDownloadResource(Resource):
```
kpsherva, line 25: This class needs a little bit untangling as well - you are using the Resource class base, which suggests you will be operating on the resource level (REST API) which you are doing when calling GET download.
However, you are using the same class to build context for UI view (not REST endpoint), which you then assign to a UI blueprint. These approaches have to be disentangled - we separate REST and UI views. Please analyse how invenio-pages module views are written (mainly REST API views, used to be mix of UI and REST) and how invenio-administration is done (mostly UI views). You can also check invenio-app-ils or cds-ils for more low level examples of UI and REST views implementation.
After applying a similar approach to those modules, the code will improve in terms of readability.

TahaKhan998: Thanks, that makes sense. I separated the download endpoint from the HTML report page now. The download resource only handles the .log download, and the administration blueprint handles the report page and its error handling. I also moved the shared run/log lookup and formatting into a neutral harvester_runs helper, so the UI page no longer depends on the download resource to build its context. This should make the REST/download part, the UI page, and the shared logic more clearly separated.

site/cds_rdm/views.py
TahaKhan998: The report page now uses a blueprint-level error handler with register_error_handler instead of returning the 404 template directly from the view.

kpsherva (approved):
nice job, thank you for cleaning up the views, looks way neater!

PR #804, opened 2026-05-20
fix(custom fields): add display URLs to CERN fields

palkerecsenyi:
The custom CERN fields used the default link (i.e. `greybook.cern.ch/{field_name}`), which resulted in broken links for nearly all of them. I have subjectively chosen relevant places to link users to for each one, and set it to `None` where there was no appropriate page to avoid having a broken link.

Feel free to suggest better destinations for any of these :)

image

### DIFF
--- a/site/cds_rdm/custom_fields/cern.py
+++ b/site/cds_rdm/custom_fields/cern.py
@@ -77,6 +77,7 @@ CERN_CUSTOM_FIELDS_UI = {
         dict(
             field="cern:committees",
             ui_widget="Dropdown",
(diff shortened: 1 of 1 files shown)


site/cds_rdm/custom_fields/cern.py
```diff
@@ -203,6 +209,7 @@
         dict(
             field="cern:beams",
             ui_widget="Dropdown",
+            display_url="https://be-dep-ea.web.cern.ch/experimental-areas/",
```
zzacharo: this has only a subset of the information right? If so, maybe omit it?

zzacharo (approved):

PR #807, opened 2026-06-04
fix(records): use header login URL in restricted files notice

### LINKED ISSUE #798: Restricted record "log in" link redirects to wrong page
Image

This log in button should redirect to the CERN SSO directly, since that's the only way to log in.

Instead, it currently redirects to this login page, which isn't accessible from anywhere else:

Image

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/798
When you're logged out and view a public record with restricted files, the "Log in" link in the red Files notice goes to /account/settings/login instead of the normal login page. This PR fixes that in detail.html by using the same login URL as the header button , so both links go to /login/?next=....

### DIFF
--- a/site/cds_rdm/templates/semantic-ui/cds_rdm/records/detail.html
+++ b/site/cds_rdm/templates/semantic-ui/cds_rdm/records/detail.html
@@ -7,6 +7,19 @@ under the terms of the GPL-2.0 License; see LICENSE file for more details.
 
 {%- extends "invenio_app_rdm/records/detail.html" %}
 
+{%- macro cds_restricted_files_access_notice(record_ui) %}
+  {%- if current_user.is_anonymous and record_ui["access"]["record"] == "public" and record_ui["access"]["files"] == "restricted" %}
+    {%- set login_url = url_for_security("login", next=request.path) -%}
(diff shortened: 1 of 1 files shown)


site/cds_rdm/templates/semantic-ui/cds_rdm/records/detail.html
```diff
@@ -41,6 +54,37 @@
 display_name=display_name) }}
 {%- endblock record_file_list -%}
 
+{%- block record_files -%}
```
kpsherva, line 57: about this one, I have a feeling that we add a lot of code to maintain. Could we have it in RDM instead so we don't need to override this big chunk of a block? it can decrease maintenance effort on CDS side (please check with f.e. @slint first if Zenodo needs this too)

TahaKhan998: After discussing with Alex, I’ve moved this fix upstream to invenio-rdm-records instead of overriding the block here. 
Upstream PR:  https://github.com/inveniosoftware/invenio-rdm-records/pull/2360
I am closing this PR.

PR #809, opened 2026-06-08
feat: per-record-version review (config stub) + update dependency major versions

palkerecsenyi:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/3403

--- 

This PR adds the config stub for per-record-version reviews. This will allow us to add a configuration that requires a review for each record version in certain circumstances (e.g. for specific communities).

**For now, the behaviour is unaffected.** `return False` means that reviews will never be required for new versions, only first versions (if the community mandates it).

I combined this with the major version upgrades since we need the new version of `invenio-rdm-records` to allow overriding the new version review policy class. **This brings CDS up to `b11.dev4`**.

### DIFF
--- a/site/cds_rdm/reviews/policy.py
+++ b/site/cds_rdm/reviews/policy.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2026 CERN.
+#
+# CDS-RDM is free software; you can redistribute it and/or modify it under
+# the terms of the GPL-2.0 License; see LICENSE file for more details.
+"""Override the policy for whether new record versions require community review."""
+
+from flask import current_app
+from invenio_rdm_records.services.review import NewRecordVersionReviewPolicy
+
+
+class CDSRecordVersionReviewPolicy(NewRecordVersionReviewPolicy):
+    """Policy override."""
+
+    @classmethod
+    def requires_review(cls, identity, draft) -> bool:
+        """Returns whether the new record version requires review."""
+        default_community = draft.parent.communities.default
+        if default_community is None:
+            return False
+
+        communities_requiring_new_record_version_review = current_app.config.get(
+            "CDS_COMMUNITIES_REQUIRING_NEW_RECORD_VERSION_REVIEW", []
+        )
+        return (
+            str(default_community.id) in communities_requiring_new_record_version_review
+        )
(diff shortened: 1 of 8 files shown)


site/cds_rdm/reviews/policy.py
```diff
@@ -0,0 +1,18 @@
...
+from invenio_rdm_records.services.review import NewRecordVersionReviewPolicy
+
+
+class CDSRecordVersionReviewPolicy(NewRecordVersionReviewPolicy):
+    """Policy override."""
+
+    @classmethod
+    def requires_review(cls, identity, draft) -> bool:
+        """Returns whether the new record version requires review."""
+        return False
```
zzacharo: maybe we need for the first testing here a config we can get the enrolled communities from so we could just configure it directly on paas instead of rebuilding the image? wdyt?

palkerecsenyi: Yes we can do, I was just going to put this stub/placeholder here so we can get the version bump merged, but I can make it decide based on a config var

palkerecsenyi: I've added the config var `CDS_COMMUNITIES_REQUIRING_NEW_RECORD_VERSION_REVIEW ` (happy to take suggestions for less verbose names hahaha) which takes a list of community UUIDs for which new-version reviews are required (if the record has that community as its default).

zzacharo (approved):

Samk13:
Question @palkerecsenyi :
Is there a specific reason for implementing this in CDS-RDM instead of supporting it in Invenio-RDM behind a config flag? 
I recall this use case being requested by the community on Discord a few times.

palkerecsenyi:
Hi @Samk13, this was implemented in InvenioRDM, see https://github.com/inveniosoftware/invenio-rdm-records/pull/2325. This PR is just enabling it for CDS according to our own requirements.

We wanted to keep the system generic enough so that any instance can specify its own set of rules for whether a new version draft should be subject to a review. So we didn't add something as explicit as "for this list of community UUIDs specifically in the config, enable new-version reviews, and disable them for everything else" directly in InvenioRDM.

This way the feature is essentially behind a config flag and isn't enabled by default so the default behaviour is kept.

Samk13:
Ah, didn't spot that on rdm-records, thanks for the clarification, great addition.
In that case, we should definitely document the new configuration pattern in the docs so others can easily adopt it.

palkerecsenyi:
Thanks! I completely agree we need to document this, I have an issue for it already https://github.com/inveniosoftware/invenio-app-rdm/issues/3425

PR #810, opened 2026-06-09
fix(harvester): improve View Changes modal UX on 403 

### LINKED ISSUE #788: Harvester Reports revision access 403 and administration errors: improve UI error handling
Revision access is based on record-level permissions instead of the global harvester curator permission.

If a curator has access to a record, they can view its revisions. If not, the API returns 403 (expected behavior).

The issue is that the Harvester Reports UI does not handle this 403 properly, so the error is unclear to the user.

Image

Update: Currently, CDS curators should not hit a 403 on revisions in normal harvester flows, because the harvester runs for a single community and all curators are owners of that community. However, this can still become an issue in the future if the harvester starts processing records for other communities, or if there are accounts with harvester admin access that are not owners/curators of a specific community. In those cases, View changes for records from that community would correctly return 403, so the UI should handle that case clearly instead of failing with a blank state.

## Related: administration error pages

When an error occurs inside the administration panel (for example 403 forbidden or 404 not found on /administration/...), users currently see the generic invenio-theme error pages. Those pages drop the admin sidebar and header, which is confusing because the user is still in the administration area.
Administration URLs should show errors inside the normal admin layout, with a clear message and a link back to the administration dashboard. Errors on the rest of the site should keep using the existing theme error pages.

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/788

- Show a clear error message in the View Changes modal when the revisions API returns 403, instead of a blank modal

image

### DIFF
--- a/site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/index.js
+++ b/site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/index.js
@@ -4,15 +4,158 @@
 // CDS-RDM is free software; you can redistribute it and/or modify it
 // under the terms of the GPL-2.0 License; see LICENSE file for more details.
 
+import React from "react";
 import { initDefaultSearchComponents } from "@js/invenio_administration";
 import { createSearchAppInit } from "@js/invenio_search_ui";
 import { NotificationController } from "@js/invenio_administration";
 import { SearchFacets } from "@js/invenio_administration";
 import { SearchBar } from "react-searchkit";
+import { ErrorMessage, withCancel } from "react-invenio-forms";
+import { Modal, Button, Loader, Grid } from "semantic-ui-react";
 import { SearchResultItemLayout } from "@js/invenio_app_rdm/administration/auditLogs/search";
 import { AuditLogActions } from "@js/invenio_app_rdm/administration/auditLogs/AuditLogActions";
+import { ViewRecentChanges } from "@js/invenio_app_rdm/administration/auditLogs/ViewRecentChanges";
+import { RecordModerationApi } from "@js/invenio_app_rdm/administration/records/api";
+import { RevisionsDiffViewer } from "@js/invenio_app_rdm/administration/components/RevisionsDiffViewer";
+import { i18next } from "@translations/invenio_app_rdm/i18next";
 import { HarvesterSearchBarElement } from "./SearchBar";
 import { CustomEmptyResults } from "./EmptyResults";
+
+function formatRevisionError(error) {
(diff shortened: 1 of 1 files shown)


site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/index.js
```diff
@@ -4,16 +4,128 @@
...
+    error.message ||
+    i18next.t("An unexpected error occurred while fetching revisions.")
+  );
+}
+
+class HarvesterAuditLogActions extends AuditLogActions {
+  onModalTriggerClick = (e, { dataActionKey }) => {
+    const { resource } = this.props;
+
+    if (dataActionKey === "view_changes") {
```
sakshamarora1: AuditLogActions should already have this buttons and the modal trigger right? I am no sure I understand why this override is needed?

site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/index.js
```diff
@@ -4,16 +4,128 @@
...
+  };
+  try {
+    await fetchPreviousRevision.call(this);
+  } finally {
+    this.setState = setState;
+  }
+};
+
+const renderViewRecentChanges = ViewRecentChanges.prototype.render;
+ViewRecentChanges.prototype.render = function () {
```
sakshamarora1: I think this can be written in a better and readable way by extending class and overriding the render method instead of injecting a new overriden render?

site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/index.js
TahaKhan998: You are right, the base class already has the buttons and opens the modal. The only reason for the override is that the base always puts the default ViewRecentChanges inside the modal and we need our own version that shows a clear message on 403. Everything else now just calls the base handler

site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/index.js
TahaKhan998: You are right, that was a bad way to do it so i changed it to a normal subclass (HarvesterViewRecentChanges extends ViewRecentChanges) that overrides render and fetchPreviousRevision.

TahaKhan998:
Moved upstream https://github.com/inveniosoftware/invenio-app-rdm/pull/3490

PR #811, opened 2026-06-10
fix(previewer): disable PDF.js streaming and range requests

TahaKhan998:
PDF previews were triggering duplicate network requests for the same file (the network tab showed the PDF being loaded twice). This adds PREVIEWER_PDF_JS_DOCUMENT_INIT_PARAMS to disable PDF.js streaming, range requests, and auto-fetch, so each preview loads the document in one request instead of multiple partial fetches against the server.

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -674,6 +674,17 @@ PREVIEWER_PREFERENCE = DEFAULT_PREVIEWER_PREFERENCE + [
   "cds_rdm_gltf_previewer",
 ]
 
+PREVIEWER_PDF_JS_DOCUMENT_INIT_PARAMS = {
+    "disableStream": True,
+    "disableRange": True,
+    "disableAutoFetch": True,
+}
+"""PDF.js initialization parameters for document previewer.
+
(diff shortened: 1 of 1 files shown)


invenio.cfg
```diff
@@ -674,6 +674,17 @@ PREVIEWER_PREFERENCE = DEFAULT_PREVIEWER_PREFERENCE + [
   "cds_rdm_gltf_previewer",
 ]
 
+PREVIEWER_PDF_JS_DOCUMENT_INIT_PARAMS = {
+    "disableStream": True,
+    "disableRange": True,
+    "disableAutoFetch": True,
```
zzacharo, line 680: curiosity: what did the autofetch do?

TahaKhan998: disableAutoFetch stops PDF.js from downloading extra bits of the PDF in the background before you actually need them like when you scroll down to the second page of the pdf, autofetch would fetch bits for the second page while you're scrolling so I turned off all three (disableRange, disableStream, disableAutoFetch) so the preview loads the file in one go instead of several small requests to the same PDF

PR #812, opened 2026-06-11
Prevent double loading on preview which triggers rate limiter

kpsherva:


zzacharo:
is this duplicate of #811 ?

kpsherva:
> is this duplicate of #811 ?

yes, thanks! #811 wasn't on the reviews so I missed it

PR #815, opened 2026-06-15
chore(setup): upgrade node dependencies

sakshamarora1:


package.json
```diff
@@ -28,11 +33,12 @@
     "i18next-browser-languagedetector": "^6.1.0",
     "jquery": "^3.3.1",
     "json-diff-kit": "^1.0.30",
+    "leaflet": "^1.9.4",
     "lodash": "^4.17.15",
     "luxon": "^1.23.0",
     "papaparse": "^5.4.1",
     "path": "^0.12.7",
-    "pdfjs-dist": "^4.0",
+    "pdfjs-dist": "^5.0",
```
palkerecsenyi, line 41: We changed this to v5 in `invenio-previewer` like [9 months ago](https://github.com/inveniosoftware/invenio-previewer/commit/ec1333d510c9acb79fee40abeb62568a47abdaa8). Does this mean we've still been using v4 on CDS since then? If this will mean a major version upgrade of pdfjs we should test it very carefully

palkerecsenyi (commented):
Looks good, we will need to do some intense testing in dev/sandbox

package.json
sakshamarora1: Looks like it. We also have to test the PDF js streaming fix

kpsherva: can you first also deploy on dev ? I didn't upgrade the frontend there

PR #821, opened 2026-06-17
feat(ep): change reportnumber generation mechanism

### LINKED ISSUE #816: EP report number minting algorithm
Check the algorithm and ensure it is correct, i.e. doesn't assign duplicate report numbers under any circumstance. Also check when we should mint the number, e.g. should it be when the request is created or only when it is approved?

zzacharo:
closes #816

### DIFF
--- a/site/cds_rdm/requests/ep_approval.py
+++ b/site/cds_rdm/requests/ep_approval.py
@@ -17,9 +17,11 @@ from flask import current_app
 from flask_principal import Identity
 from invenio_access.permissions import system_identity
 from invenio_db import db
+from invenio_drafts_resources.services.records.uow import ParentRecordCommitOp
 from invenio_i18n import lazy_gettext as _
 from invenio_notifications.services.uow import NotificationOp
 from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+from invenio_rdm_records.proxies import current_rdm_records_service
 from invenio_rdm_records.records.api import RDMRecord
 from invenio_rdm_records.requests.base import BaseRequest as RDMBaseRequest
 from invenio_records_resources.services.errors import PermissionDeniedError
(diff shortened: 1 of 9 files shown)


site/cds_rdm/requests/ep_approval.py
```diff
@@ -157,47 +159,61 @@ def _community_config(self):
...
+        digits = rn.get("counter_digits", 3)
         year = date.today().year
-        # Split on "{seq" to get everything before the sequence placeholder,
-        # then format only the year part → "CERN-EP-2026-"
-        # TODO: I don't understand what this is doing
-        prefix = pattern.split("{seq")[0].format(year=year)
+        scan_prefix = f"{prefix}-{year}-" if rn.get("include_year", True) else f"{prefix}-"
+
+        lock_key = int.from_bytes(
+            hashlib.sha256(scan_prefix.encode()).digest()[:8], "big", signed=True
```
palkerecsenyi: I understand what this is doing after reading it carefully, but it looks quite complex. Maybe could you include a short comment to explain + a link to the Postgres docs on advisory locks? Since these are quite a niche feature.

site/cds_rdm/requests/ep_approval.py
```diff
@@ -228,11 +239,13 @@ def execute(self, identity: Identity, uow: UnitOfWork) -> None:
...
-        # but why not create a generic `metadata` field or something.
         topic.parent["permission_flags"] = pf
-        topic.parent.commit()
-        db.session.commit()
+
+        # Flush the parent and defer the DB commit to the UoW so that the
+        # parent write, the PID insert, and the request status update all land
+        # in the same transaction.  The advisory lock acquired in
+        # _issue_report_number is held until this transaction commits.
+        uow.register(ParentRecordCommitOp(topic.parent))
```
palkerecsenyi: Do we not need to pass the indexer into the commit op?

palkerecsenyi (approved):
Looks good, just some minor things!

site/cds_rdm/requests/ep_approval.py
zzacharo: I removed the lock and let's reevaluate if needed the solution

palkerecsenyi (approved):

PR #831, opened 2026-06-26
Improve harvester log grouping and error reporting

### LINKED ISSUE #813: harvester: error log without raising to sentry
Image

The errors (Skipping entry) are not handled correctly - it should show as an error in the logs of the harvester but not be reported in sentry (since it will have a lot of exceptions potentially) - at least, if reporting to sentry is unavoidable, we should raise it in a way that sentry groups them by default

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/813
This PR improves INSPIRE harvester run error reporting by turning errors and warning into clear grouped errors and warnings, making log output much more useful for curators, and ensuring the report highlights the real failure reasons instead of vague or repetitive raw lines. It adds structured grouping for repeated failures and warnings, improves the plain-text download output to match that grouped view, and updates harvester writer messages so record-level failures explain what went wrong in a more actionable way, including which records were affected and why they failed.

https://github.com/user-attachments/assets/2eae86ca-208c-4f9d-8358-d6feabe17ef6

This is what a downloaded file of harvester logs looks like now:

[harvester_logs_4a7fc391-5b9a-45b3-873f-bd2d2a297aaa_20260626_120105.log](https://github.com/user-attachments/files/29377317/harvester_logs_4a7fc391-5b9a-45b3-873f-bd2d2a297aaa_20260626_120105.log)


site/cds_rdm/harvester_runs/logs.py
```diff
@@ -33,18 +41,525 @@ def __init__(self, message, code):
 
 def format_timestamp(value):
     """Format timestamps for display."""
-    if value is None or value == "":
+    if value in (None, ""):
```
kpsherva:
```suggestion
    if not value:
```

site/cds_rdm/harvester_runs/logs.py
```diff
@@ -33,18 +41,525 @@ def __init__(self, message, code):
...
+        except (TypeError, ValueError):
             return str(value)
     return format_datetime(dt, "yyyy-MM-dd HH:mm")
 
 
+# String helpers -------------------------------------------------------------
+
+def _compact(value):
+    """Collapse whitespace while keeping the original text readable."""
+    text = str(value or "").replace("\\n", " ").replace("\n", " ")
```
kpsherva: where are the whitespaces originally added?

site/cds_rdm/harvester_runs/logs.py
```diff
@@ -33,18 +41,525 @@ def __init__(self, message, code):
...
+    return " ".join(text.split()).strip()
+
+
+def _remove_prefix(text, prefix):
```
kpsherva: which prefixes are removed?

site/cds_rdm/harvester_runs/logs.py
```diff
@@ -33,18 +41,525 @@ def __init__(self, message, code):
...
+        parsed = ast.literal_eval(text)
+    except (ValueError, SyntaxError):
+        return None
+    if isinstance(parsed, dict):
+        value = parsed.get("id")
+        return str(value) if value not in (None, "") else None
+    return None
+
+
+def _strip_leading_bracket_prefixes(message):
```
kpsherva: there is a lot of things happening here which I need some narration from your side to understand. Please book (on calendar) some time for us to go through the code together

site/cds_rdm/harvester_runs/logs.py
```diff
@@ -84,33 +186,248 @@ def fetch_harvester_run_logs(run):
...
 
-def lines_from_hits(hits):
-    """Return de-duplicated log lines and severity counts."""
-    task_groups = {}
+def _extract_record_id(message):
+    match = _RECORD_ID_PATTERN.search(message or "")
+    return match.group("id") if match else None
+
+
+def _unwrap_skipped_entry_error(reason):
```
TahaKhan998: remove dictonary

site/cds_rdm/harvester_runs/logs.py
```diff
@@ -84,33 +186,248 @@ def fetch_harvester_run_logs(run):
...
+
+def _is_skipped_entry_log(message):
+    text = compact_text(message)
+    return text.startswith("Skipped entry ") or (
+        text.startswith("Skipping ")
+        and text.endswith(" transformed entries with errors.")
+    )
+
+
+def _safe_strip_wrapper_prefixes(reason):
```
kpsherva: check if this can be fixed while logging in the first place - are the classes added in the harvester task? if yes we can reorganise there instead of parsing. If these classes are added in the jobs - let's discuss

site/cds_rdm/harvester_runs/logs.py
```diff
@@ -84,33 +186,248 @@ def fetch_harvester_run_logs(run):
...
+
+
+def _finalize_group_title(title):
+    """Never expose bare payload placeholders as curator-facing titles."""
+    stripped = compact_text(title)
+    if not stripped:
+        return "Unknown error"
+
+    if _PLACEHOLDER_ONLY_TITLE.match(stripped):
+        return "Record validation failed." if stripped == "<payload>" else "Log message"
```
kpsherva: please check how to fix `..` errors and display the error correctly instead of stripping the payload

site/cds_rdm/harvester_runs/logs.py
```diff
@@ -84,33 +186,248 @@ def fetch_harvester_run_logs(run):
...
+    lower = reason.lower()
+    if any(hint in lower for hint in _VALIDATION_HINTS):
+        return "Record validation failed."
+    if "must be published in the" in lower and "community" in lower:
+        return "Record validation failed."
+    return None
+
+
+def _known_error_stem(reason):
+    """Normalize legacy and current variants of recurring harvester errors."""
```
kpsherva: is this normalisation needed?

site/cds_rdm/inspire_harvester/reports/runs/logs.py
```diff
@@ -66,51 +94,294 @@ def resolve_harvester_run(run_id):
...
+    """
+    text = compact_text(message)
+    if text.startswith("Skipping ") and text.endswith(_SKIP_SUMMARY_SUFFIX):
+        return "summary"
+    if text.startswith("Skipped entry "):
+        return "entry"
+    return None
+
+
+def _peel_string_list(text):
```
kpsherva, line 171: what is the case for which we are returning a list in the log? maybe the log is incorrect?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -30,11 +29,64 @@
...
     UpdateEngineConflict,
 )
+from cds_rdm.utils import compact_text, flatten_error_payload
 
 
 class InspireWriter(BaseWriter):
     """INSPIRE writer."""
 
+    @staticmethod
+    def _format_validation_error(error):
```
kpsherva: these methods would fit better in the logger.py, as logging is not the responsibility of this class. What do you think?

site/cds_rdm/harvester_download/resources/resource.py
```diff
@@ -16,7 +16,7 @@
 from cds_rdm.harvester_runs.logs import (
```
kpsherva: feedback not connected with this file but with the directory:
could you move harvester_download and harvester_runs into `inspire_harvester/reports/download` and `inspire_harvester/reports/runs` - to package the whole functionality of the harvester together

site/cds_rdm/inspire_harvester/reports/runs/logs.py
```diff
@@ -66,51 +94,294 @@ def resolve_harvester_run(run_id):
...
-        hits = list(result.hits)
-        total = result.total or len(hits)
+        page_size = current_app.config.get("JOBS_LOGS_MAX_RESULTS", 2000)
+
+        for _ in range(max_pages):
+            params = {"q": f'"{run.id}"', "sort": "timestamp"}
+            if search_after:
+                params["search_after"] = search_after
+
+            result = current_jobs_logs_service.search(system_identity, params=params)
```
kpsherva, line 110: .search function will return only 10k results max, even with the pagination

site/cds_rdm/utils.py
```diff
@@ -14,6 +14,49 @@
 from invenio_records_resources.services.errors import ValidationError
 
 
+def compact_text(value):
+    """Collapse whitespace (including newlines) into single spaces."""
+    text = str(value or "").replace("\\n", " ").replace("\n", " ")
+    return " ".join(text.split()).strip()
+
+
+def flatten_error_payload(value, prefix=""):
```
kpsherva: I think there is a function which does something very similar: search for validation_error_to_list_errors/_iter_errors_dict in invenio_records_resources.errors

site/cds_rdm/inspire_harvester/reports/runs/logs.py
TahaKhan998: Good question. The screenshot shows where this comes from,  in invenio-vocabularies, when a record fails it logs result.errors directly, and that's a list. So what ends up in the log looks like ['error message here'] instead of plain text. _peel_string_list just strips that wrapper off so the report can group errors properly. For example, without it we'd group on something like skipped entry with errors: ['more than 1 doi was found.'], brackets, quotes and all. After peeling we group on just more than 1 doi was found., so records that failed for the same reason actually end up in the same group.

It would be nicer if invenio-vocabularies wrote it as plain text in the first place. I handled it in the report for now since past harvester runs already have logs saved this way, but if you'd rather fix it at the source I can do that instead.

Image

site/cds_rdm/inspire_harvester/reports/runs/logs.py
TahaKhan998: In invenio-jobs it’s 2000 per .search() by default (screenshot). Not sure about the 10k,  maybe that’s in the CDS deployment config? Our loop just keeps fetching batches of 2000 with search_after until we’ve got all the logs or hit the page limit. The old code only fetched once.

Image

site/cds_rdm/harvester_download/resources/resource.py
TahaKhan998: Done

site/cds_rdm/inspire_harvester/writer.py
TahaKhan998: You're right — moved them to logger.py.

site/cds_rdm/utils.py
TahaKhan998: Yep, switched to the invenio helper and dropped the duplicate.

site/cds_rdm/inspire_harvester/reports/runs/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2026 CERN.
+#
+# CDS-RDM is free software; you can redistribute it and/or modify it
+# under the terms of the GPL-2.0 License; see LICENSE file for more details.
```
kpsherva: GPL license is not correct, to be changes in all the occurences

site/cds_rdm/inspire_harvester/reports/runs/logs.py
kpsherva: 10k is the opensearch limit on .search interface. I didn't remember we had the JOB config too... 
but in this case, which limit is applied? 10k or 2k?

TahaKhan998: I checked this again. The 10k limit is the OpenSearch limit for normal from/size pagination, but that is not what we use here. invenio-jobs uses search_after and returns a maximum of 2,000 logs per .search() call, fetching them internally in batches of 500. Our loop then uses the last result to make another .search() call and fetch the next 2,000 logs. Therefore, the OpenSearch 10k limit does not apply here. The applied limit is 2,000 logs per .search() call, and since our code can make up to 50 calls, the overall limit is 100,000 logs.

site/cds_rdm/inspire_harvester/reports/runs/__init__.py
TahaKhan998: Changed in all the occurrences now.

PR #832, opened 2026-06-26
feat(assets): Switch to pnpm and rspack

sakshamarora1:


site/cds_rdm/assets/semantic-ui/js/cds_rdm/administration/harvesterReports/SearchBar.js
```diff
@@ -63,7 +70,7 @@ const SearchBarComponent = ({ updateQueryState, currentQueryState }) => {
       setActiveRunId(defaultRun.id);
       executeSearch(defaultRun, "");
     }
-  }, []);
+  }, [currentQueryState, runs, defaultRun]);
```
kpsherva, line 77: could you check this warning?

site/.npmrc
```diff
@@ -0,0 +1 @@
+lockfile=false
```
kpsherva, line 1: what will this do?

pnpm-lock.yaml
```diff
@@ -0,0 +1,12382 @@
+lockfileVersion: '9.0'
```
kpsherva, line 1: this is quite sensitive change. If I remember correctly package-lock.json is used to build our deployment images - did you adapt that to the change, will the image still build correctly?

site/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "cds-rdm-site",
```
kpsherva: why do we need to add package.json here?

.github/workflows/tests.yml
```diff
@@ -74,10 +74,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v5
 
-      - name: Use Node.js v20.x
-        uses: actions/setup-node@v6
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
```
palkerecsenyi: Any reason why we are not using v6 of the pnpm action?

site/package.json
palkerecsenyi: I am also confused about this, it would maybe make sense if it was included as a `workspace` in the root level `package.json` and all of the dependencies were instead listed here. But I'm not sure why it's necessary in the current form.

site/.npmrc
sakshamarora1: Disable package-lock.json

pnpm-lock.yaml
sakshamarora1: Yes, this is also how Zenodo and other instances are doing it as well.

site/package.json
sakshamarora1: It is to fix the broken linter, purpose mentioned in the description already. For the dependencies for eslint.

.github/workflows/tests.yml
sakshamarora1: No reason 👀 
I will update

Dockerfile
```diff
@@ -81,13 +99,18 @@ RUN --mount=type=cache,target=/opt/.cache/uv \
...
+RUN --mount=type=cache,target=/opt/.cache/pnpm-store \
+    invenio collect --verbose && \
+    invenio webpack create && \
+    CI=true invenio webpack install
 
-# Install JS deps from the package-lock file
-COPY package-lock.json ${INVENIO_INSTANCE_PATH}/assets/
+COPY ./assets/ ${INVENIO_INSTANCE_PATH}/assets/
+COPY ./static/ ${INVENIO_INSTANCE_PATH}/static/
+RUN cd ${INVENIO_INSTANCE_PATH}/assets && pnpm run build
```
kpsherva, line 112: could you keep it the way it was? that we copy the file from the repo to instance assets and we use that to install later?

sakshamarora1: Yep it's doing that. See above at line 51
Intentionally moved above the python COPYs so it caches. Moving it below webpack create is the same but reinstalls JS deps on every rebuild

PR #835, opened 2026-06-29
Ep approval rename

zzacharo:


site/cds_rdm/ext.py
```diff
@@ -16,13 +18,33 @@
...
 )
-from cds_rdm.requests.ep_approval_state import get_ep_approval_state
+from cds_rdm.requests.committee_approval_state import get_committee_approval_state
 
 from . import config
 from .utils import evaluate_permissions
 from .views import get_linked_records_search_query
 
 
+class _StripCommitteeApprovalDumperExt(SearchDumperExt):
```
zzacharo: @palkerecsenyi here I decided to strip out the field from being indexed as I had troubles making my mind regarding the mapping and I thought to keep it only in DB for now before we get some feedback for the needed searches.

site/cds_rdm/ext.py
```diff
@@ -39,17 +61,24 @@ def init_config(self, app):
...
         app.jinja_env.filters["get_linked_records_search_query"] = (
             get_linked_records_search_query
         )
         return app
 
+    def _register_dumper_extensions(self):
+        """Register CDS-specific dumper extensions on InvenioRDM record models."""
+        from invenio_rdm_records.records.api import RDMParent
+
+        RDMParent.dumper._extensions.append(_StripCommitteeApprovalDumperExt())
```
zzacharo: This needs maybe some <3 on how to properly do it for RDM

palkerecsenyi:
note to self: check the PID indexing in the database to ensure the `LIKE` query will scale without problems

edit: not a problem since we use `pid_type = 'apprn'` in conjunction with the `pid_value LIKE ...`. So it will use the binary tree unique index (on `pid_type` and `pid_value`) to _first_ very quickly filter only the `apprn` PIDs, and then more slowly row-by-row check the `LIKE` queries. But we don't expect a very large number of such PIDs, so performance will not really be an issue.

palkerecsenyi (approved):

PR #836, opened 2026-06-30
feat: add ATOM related identifier scheme

### LINKED ISSUE #786: Add ATOM ID scheme in related identifiers
decided during the meeting that we need an "ARCHIVE" scheme for the related identifiers - not atom ID since ATOM does not generate these

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/786 
 
Adds ATOM as a CDS-specific related identifier scheme so it can be selected in the Related works section like the other CDS-only schemes. The change keeps the implementation consistent with the existing project pattern by registering the scheme in CDS config, adding the validator/normalizer in schemes.py, and updating the test config accordingly.

### DIFF
--- a/site/cds_rdm/schemes.py
+++ b/site/cds_rdm/schemes.py
@@ -27,6 +27,7 @@ hal_regexp = re.compile(r"^(hal:|HAL:)?((in2p3|[a-z]{3}[a-z]*)-|(sic|mem|ijn)_)\
 # Matches patterns like CERN-EP-2026-001 or CERN-TH-2026-042, as well as
 # legacy multi-segment department codes like CERN-PH-EP-2012-369.
 approval_rn_regexp = re.compile(r"^[A-Z]+(?:-[A-Z]+)*-\d{4}-\d+$")
+archive_pattern = re.compile(r"^CERN-ARCH(?:-[A-Z0-9]+)+$", flags=re.I)
 
 
 def is_aleph(val):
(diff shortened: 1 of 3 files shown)


site/tests/conftest.py
```diff
@@ -236,6 +236,11 @@ def app_config(app_config, mock_datacite_client, mock_crossref_client):
             "validator": schemes.is_edms,
             "datacite": "EDMS",
         },
+        "atom": {
```
sakshamarora1: Do we need this in the tests? I don't see any tests for it, and also do we need any commands?

TahaKhan998: I checked the other CDS-specific related identifiers, and some of them also do not have dedicated tests, so I don’t think ATOM needs its own test either so I removed the changes in conftest.py.  No commands, migrations, or reindexing are needed for this change.

site/cds_rdm/schemes.py
```diff
@@ -26,6 +26,11 @@
 edms_pattern = re.compile(r"^\d+$", flags=re.I)
 
 
+def is_atom(val):
+    """Test if argument is a non-empty ATOM identifier."""
+    return bool(str(val).strip())
```
kpsherva: how do you know what format atom ID has?

palkerecsenyi (approved):

zzacharo (approved):

PR #837, opened 2026-06-30
feat(components): Auto submit public records to cern research

### LINKED ISSUE #830: submit automatic inclusion request to CERN Scientific community for submissions of public research type records
UPDATE: it seems that the inclusion request to CERN Research is sent on each publish - it happened when we bulk updated metadata on FASER collection. It needs to be tested against creating too many requests

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-rdm/issues/830

### DIFF
--- a/invenio.cfg
+++ b/invenio.cfg
@@ -82,10 +82,7 @@ from invenio_app_rdm.config import \
 from invenio_app_rdm.config import \
     VOCABULARIES_DATASTREAM_WRITERS as DEFAULT_VOCABULARIES_DATASTREAM_WRITERS
 from cds_rdm.clc_sync.services.components import ClcSyncComponent
-from cds_rdm.components import CDSResourcePublication
-from cds_rdm.components import CommitteeApprovalComponent
-from cds_rdm.components import SubjectsValidationComponent
-from cds_rdm.components import MintAlternateIdentifierComponent
+from cds_rdm.components import CDSResourcePublication, CommitteeApprovalComponent, MintAlternateIdentifierComponent, PublicationInclusionComponent, SubjectsValidationComponent
 from cds_rdm.pids import validate_optional_doi_transitions
 from cds_rdm.views import frontpage_view_function, inspire_link_render
 from cds_rdm.reviews.policy import CDSRecordVersionReviewPolicy
@@ -577,6 +574,26 @@ of 30 days after first publication.
 E.g. if the bucket is unlocked 29 days after initial publication, files can only be modified for 1 day.
 """
 
+# Public research publication resource types that should be included in the CERN Scientific community.
+CDS_CERN_SCIENTIFIC_RESOURCE_TYPES = {
+    "publication-dissertation",
+    "publication-book",
+    "publication-section",
+    "publication-conferencepaper",
+    "publication-conferenceproceeding",
+    "publication-conferencenote",
+    "publication-journal",
+    "publication-article",
+    "publication-preprint",
+    "publication-report",
+    "publication-technicalnote",
+    "publication-note",
+    "publication",
+}
+"""
+Resource types that should be included in the CERN Scientific community.
+"""
+
 RDM_SEARCH = {
     **deepcopy(RDM_SEARCH),
     "query_parser_cls": QueryParser.factory(
@@ -740,6 +757,7 @@ RDM_RECORDS_SERVICE_COMPONENTS = [
     CDSResourcePublication,
     ClcSyncComponent,
     MintAlternateIdentifierComponent,
+    PublicationInclusionComponent,
     VCSComponent,
 ]
 
--- a/site/cds_rdm/components.py
+++ b/site/cds_rdm/components.py
@@ -21,23 +21,7 @@ from invenio_rdm_records.services.errors import ValidationErrorWithMessageAsList
 from invenio_records_resources.services.uow import TaskOp
 from marshmallow import ValidationError
 
-from .tasks import sync_alternate_identifiers
-
-# @shared_task()
-# def create_community_inclusion_request(record_id):
-#     """Create a community inclusion request for ."""
-#     # Create a community-inclusion request
-#     csc_community_id = current_app.config.get("CDS_CERN_SCIENTIFIC_COMMUNITY_ID")
-#     data = dict(
-#         communities=[
-#             {
-#                 "id": csc_community_id,
-#                 "require_review": True,
-#             }
-#         ]
-#     )
-
-#     current_rdm_records.record_communities_service.add(system_identity, record_id, data)
+from .tasks import submit_community_inclusion_request, sync_alternate_identifiers
 
 
 def is_record_public(record):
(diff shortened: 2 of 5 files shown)


site/cds_rdm/components.py
```diff
@@ -218,3 +202,50 @@ def publish(self, identity, draft=None, record=None):
...
+    after the publish transaction commits.
+    """
+
+    def _is_eligible(self, draft, record):
+        """Return True when the record should be auto-submitted for community inclusion."""
+        if not is_record_public(draft):
+            return False
+
+        resource_type = draft["metadata"]["resource_type"]["id"]
+        public_research_types = current_app.config.get(
```
kpsherva:
```suggestion
        research_types = current_app.config.get(
```
this is a bit confusing as a variable name. Do we have anything like public research types? did you mean research resource types?

site/cds_rdm/components.py
```diff
@@ -218,3 +202,50 @@ def publish(self, identity, draft=None, record=None):
...
+            "CDS_PUBLIC_RESEARCH_RESOURCE_TYPES", set()
+        )
+        if resource_type not in public_research_types:
+            return False
+
+        csc_community_id = current_app.config.get("CDS_CERN_SCIENTIFIC_COMMUNITY_ID")
+        if not csc_community_id:
+            current_app.logger.warning(
```
kpsherva: this should be an error not a warning and we want this popping up sentry as an exception, this variable is very important and if it is not configured, we should know

site/cds_rdm/components.py
```diff
@@ -218,3 +202,50 @@ def publish(self, identity, draft=None, record=None):
...
+                record.pid.pid_value,
+            )
+            return False
+
+        if csc_community_id in record.parent.communities.ids:
+            return False
+
+        return True
+
+    def publish(self, identity, draft=None, record=None, **kwargs):
```
kpsherva, line 386: do we auto accept it? or it needs to be accepted?

site/cds_rdm/tasks.py
```diff
@@ -312,3 +313,58 @@ def sync_alternate_identifiers(parent_id, record_id):
...
             db.session.add(report_number_pid)
     db.session.commit()
+
+
+@shared_task(ignore_result=True)
+def submit_community_inclusion_request(record_id):
+    """Create and submit a community inclusion request to the CERN Scientific Community after a public record satisfies the criteria."""
+    csc_community_id = current_app.config.get("CDS_CERN_SCIENTIFIC_COMMUNITY_ID")
+    if not csc_community_id:
+        current_app.logger.warning(
```
kpsherva: same as above, also since the task is checking it here, do we need to do the same in the component? isn't it redundant in the component?

site/cds_rdm/tasks.py
```diff
@@ -312,3 +313,58 @@ def sync_alternate_identifiers(parent_id, record_id):
...
+        return
+
+    data = {
+        "communities": [
+            {
+                "id": csc_community_id,
+                "require_review": True,
+                "comment": {
+                    "payload": {
+                        "content": "<p>This record was automatically submitted to the CERN Research Community because it satisfies the criteria for inclusion.</p>",
```
kpsherva:
```suggestion
                        "content": "<p>This record was automatically submitted to the CERN Research Community by the system because it satisfies the criteria for inclusion (resource type identified as research related) </p>",
```

site/cds_rdm/tasks.py
```diff
@@ -312,3 +313,58 @@ def sync_alternate_identifiers(parent_id, record_id):
...
+                        "format": "html",
+                    }
+                },
+            }
+        ]
+    }
+
+    try:
+        with UnitOfWork() as uow:
+            processed, errors = current_record_communities_service.add(
```
kpsherva: doesn't it automatically add the community to the record without requiring the review? We discussed that initially we should go through approval, we might switch to automatic approval later on

sakshamarora1: If that happens, later on we can remove the require_review flag from the data being passed to this service method.

site/cds_rdm/components.py
sakshamarora1: In the task, there is [`"require_review": True`](https://github.com/CERNDocumentServer/cds-rdm/pull/837/files#diff-48d7dfeba7dedee600381f6c5491916254d518942b2c2708f42135df63e3a690R334)
It only submits for review

site/cds_rdm/components.py
sakshamarora1: Alright, updated!

site/cds_rdm/components.py
```diff
@@ -21,23 +21,7 @@
 from invenio_records_resources.services.uow import TaskOp
 from marshmallow import ValidationError
 
-from .tasks import sync_alternate_identifiers
-
-# @shared_task()
-# def create_community_inclusion_request(record_id):
```
zzacharo, line 27: @sakshamarora1 I had this from the initial days of our first deployment. Maybe delete or use it?

site/cds_rdm/components.py
```diff
@@ -363,3 +347,50 @@ def publish(self, identity, draft=None, record=None):
                 record_id=str(record.id),
             )
         )
+
+
+class PublicationInclusionComponent(ServiceComponent):
```
zzacharo, line 352: this is yet another way to include to CERN research? For thesis we handle it differently or it will incorporate also that case?

site/cds_rdm/tasks.py
```diff
@@ -312,3 +313,51 @@ def sync_alternate_identifiers(parent_id, record_id):
...
+@shared_task(ignore_result=True)
+def submit_community_inclusion_request(record_id):
+    """Create and submit a community inclusion request to the CERN Scientific Community after a public record satisfies the criteria."""
+    csc_community_id = current_app.config.get("CDS_CERN_SCIENTIFIC_COMMUNITY_ID")
+
+    data = {
+        "communities": [
+            {
+                "id": csc_community_id,
+                "require_review": True,
```
zzacharo, line 327: Btw for EP [here](https://github.com/CERNDocumentServer/cds-rdm/blob/c84ba358c05b4a90f0c72bb20997a458d1a4248c/site/cds_rdm/requests/views.py#L254) we auto accept the records but maybe is fine for that case as it is more curated content @kpsherva

site/cds_rdm/tasks.py
kpsherva: I am confused from all the questions and feedback - is the inclusion request to CERN research auto accepted or not in the end?

site/cds_rdm/tasks.py
```diff
@@ -312,3 +313,51 @@ def sync_alternate_identifiers(parent_id, record_id):
...
+    try:
+        with UnitOfWork() as uow:
+            processed, errors = current_record_communities_service.add(
+                system_identity, record_id, data, uow=uow
+            )
+            uow.commit()
+
+        if errors:
+            for error in errors:
+                current_app.logger.info(
```
kpsherva: this shouldn't be info... we should log it with an error and get sentry exception, no?

site/cds_rdm/tasks.py
```diff
@@ -312,3 +313,51 @@ def sync_alternate_identifiers(parent_id, record_id):
...
+                    error.get("message"),
+                )
+        else:
+            current_app.logger.info(
+                "Community inclusion request %s submitted for record %s to community %s.",
+                processed[0].get("request_id"),
+                record_id,
+                processed[0].get("community_id"),
+            )
+    except Exception as e:
```
kpsherva: what other errors can we get? are you sure the error message is serialized correctly in the f-string and will be reported correctly?

invenio.cfg
```diff
@@ -577,6 +574,25 @@ of 30 days after first publication.
...
+    "publication-book",
+    "publication-section",
+    "publication-conferencepaper",
+    "publication-conferenceproceeding",
+    "publication-journal",
+    "publication-article",
+    "publication-preprint",
+    "publication-report",
+    "publication-technicalnote",
+    "publication-note",
```
kpsherva, line 590:
```suggestion
    "publication-note",
    "publication-conferencenote",
```

which was recently added

site/cds_rdm/tasks.py
sakshamarora1: [Not auto accepted](https://github.com/CERNDocumentServer/cds-rdm/pull/837/files#diff-48d7dfeba7dedee600381f6c5491916254d518942b2c2708f42135df63e3a690R327)

site/cds_rdm/tasks.py
sakshamarora1: I didn't find any in my testing. If it's not needed, can be removed, let me know

sakshamarora1: removed

zzacharo (approved):

PR #841, opened 2026-07-06
EP bug fixes + warning messages

### LINKED ISSUE #840: EP approval flow bug fixes
- [x] Check why record is not being submitted to scientific community on final publish
- [x] Check permissions for the "Original record" link (it is e.g. not being shown if the user is a superadmin (and therefore has edit permissions on the record) but does not have a curator+ role in the community)
  Image
- [x] If the original record has share access, is this transferred over to the public record?
  - We should not transfer the permissions to keep it simple. Users can manually add share access to the final public record if they need.
  - Edit: no it is not transferred, the new record is created without any access grants. The files and metadata are both set to public viewable (regardless of the original record's settings) and no grants are copied over.

### LINKED ISSUE #842: Record management warnings for EP
Add a warning when clicking the new version button on a record that has a pending EP approval request (i.e. has not yet been accepted/rejected)

Also add a warning when trying to submit an approval request for a non-restricted (public) record.

palkerecsenyi:
Closes #840 

Closes https://github.com/CERNDocumentServer/cds-rdm/issues/842

**Depends on https://github.com/inveniosoftware/invenio-rdm-records/pull/2379, please merge+release this before deploying this PR.**

---

* Changed permissions of the "original record" link to always show if the user has read access to the original draft record. This involves an additional DB/OS query, so I made sure it only runs if the manage menu renders (i.e. the user is able to edit the final public record).

* Added an ignorable warning when the user is trying to create a new record version after a committee approval request has already been created (regardless of the status of the request).

* The final public record is now automatically included in the CERN Research community. The inclusion request is still created but it is automatically accepted and requires no manual approval.

* Added an ignorable warning when the committee approval request is being created for a record when that record or its files are not restricted.


assets/js/components/record_details/NewVersionButton.js
```diff
@@ -0,0 +1,114 @@
...
+
+      <Modal open={showModal} onClose={() => setShowModal(false)} size="small">
+        <Modal.Header>
+          <Icon name="warning sign" color="yellow" className="mr-10" />
+          {i18next.t("EP approval request pending")}
+        </Modal.Header>
+        <Modal.Content>
+          <p>
+            {i18next.t(
+              "An EP approval request is already pending for an existing version of this record. " +
```
zzacharo, line 82: This should be fine at this point but we should configure these places or start with generic terms like `Committee approval request pending`

palkerecsenyi: I have created an issue in the backlog to genericise all the UI labels, there are a couple others that need changing too. I guess we can leave this shelved for now until we have another committee approval use case or we have some time to implement it

https://github.com/CERNDocumentServer/cds-rdm/issues/843

zzacharo (approved):

PR #844, opened 2026-07-09
change(custom_fields): conference as an array

kpsherva:


site/cds_rdm/custom_fields/meeting.py
```diff
@@ -20,7 +90,7 @@
         {
             "field": "meeting:meeting",
             "ui_widget": "CDSMeeting",
-            "template": "meeting.html",
+            "template": "CDSMeeting.html",
             "props": {
```
zubeydecivelek, line 96: Should we need to add `url` here? it's defined [here](https://github.com/CERNDocumentServer/cds-rdm/pull/844/changes#diff-71c8e33caebc2f87853eaaffc0336a8a24dfd16a1d463c5cd32bef1b822118c5R40)
Like:
```python
"url": {
    "label": _("Website"),
    "placeholder": "",
    "description": "",
},
```

kpsherva: thanks! this is incorrect in fact, because we have URLs in the identifiers... good catch

kpsherva: actually, it is even worse. The URL is backwards compatibility field, a leftover from the time when we had an URL.

PR #845, opened 2026-07-09
fix(ep): clarify wording in new version warning

palkerecsenyi:
The warning is shown whenever a Committee Approval request exists for (any version of) the record, regardless of what status the request has. Even if it is accepted, the warning will still show. So I have changed the message slightly to avoid confusion for non-pending statuses.

### DIFF
--- a/assets/js/components/record_details/NewVersionButton.js
+++ b/assets/js/components/record_details/NewVersionButton.js
@@ -14,15 +14,19 @@ export const NewVersionButton = ({ onError, record, disabled, ...uiProps }) => {
   const [loading, setLoading] = useState(false);
   const [showModal, setShowModal] = useState(false);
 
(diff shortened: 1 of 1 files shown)


assets/js/components/record_details/NewVersionButton.js
```diff
@@ -79,13 +79,13 @@ export const NewVersionButton = ({ onError, record, disabled, ...uiProps }) => {
         <Modal.Content>
           <p>
             {i18next.t(
-              "An EP approval request is already pending for an existing version of this record. " +
+              "There is already an EP approval request for an existing version of this record. " +
```
zzacharo: if there is no pending request we should not show the warning.

zzacharo (approved):

PR #852, opened 2026-07-13
feat: genericise EP approval UI labels via config

### LINKED ISSUE #843: Genericise EP approval UI labels
We want to allow the term "EP" to be configurable (via `invenio.cfg`) wherever it is used. E.g. instead of `"EP Board review"` we should use `"${{committee_name}} review"` or something like this. Where using a variable is too complicated, we can use the generic term "Committee review".

The terms have already been genericised in the code (i.e. variable and method names, etc).

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/843

This PR makes the EP approval UI labels configurable from CDS_COMMITTEE_APPROVAL_COMMUNITIES in invenio.cfg instead of hardcoding EP-specific wording everywhere. I used the existing label config for EP approval-style text, added optional committee_name support for review-step wording like EP Board review, and kept generic committee wording in places where direct substitution would be awkward. I also cleaned up the implementation so the workflow config is resolved through the shared helper and used consistently across the record page, request flow, and submit-notification email.

### DIFF
--- a/assets/js/components/record_details/CommitteeApproval.js
+++ b/assets/js/components/record_details/CommitteeApproval.js
@@ -65,6 +65,7 @@ export class CommitteeApprovalManageSection extends Component {
     // Public EP-approved record — show a compact provenance note.
     if (committeeApproval.is_public_approved_record) {
       const {
+        approval_label: approvalLabel,
         approved_report_number: pubRn,
         draft_record_id: draftRecordId,
         can_view_reviewed_version: canViewReviewedVersion,
@@ -73,7 +74,11 @@ export class CommitteeApprovalManageSection extends Component {
       return (
         <Grid.Column className="pb-20 pt-0">
           <Divider horizontal>
-            <Header as="h4">{i18next.t("Approval request workflow")}</Header>
+            <Header as="h4">
+              {i18next.t("{{approvalLabel}} request workflow", {
+                approvalLabel,
+              })}
+            </Header>
           </Divider>
 
           <p className="text-muted text-align-center">
@@ -104,6 +109,8 @@ export class CommitteeApprovalManageSection extends Component {
     const {
       can_submit: canSubmit,
       can_create_public: canCreatePublicFlag,
+      approval_label: approvalLabel,
+      committee_name: committeeName,
       open_request: openRequest,
       approved_report_number: approvedReportNumber,
       receiver_group: receiverGroup,
@@ -151,7 +158,11 @@ export class CommitteeApprovalManageSection extends Component {
     return (
       <Grid.Column className="pb-20 pt-0">
         <Divider horizontal>
-          <Header as="h4">{i18next.t("Approval request workflow")}</Header>
+          <Header as="h4">
+            {i18next.t("{{approvalLabel}} request workflow", {
+              approvalLabel,
+            })}
+          </Header>
         </Divider>
 
         <Step.Group
@@ -191,7 +202,12 @@ export class CommitteeApprovalManageSection extends Component {
                     ) : step1Completed ? (
                       i18next.t("Request submitted.")
                     ) : (
-                      i18next.t("Submit the document for EP committee review.")
+                      i18next.t(
+                        "Submit the document for {{committeeName}} review.",
+                        {
+                          committeeName,
+                        }
+                      )
                     )}
                   </Step.Description>
                 </div>
@@ -228,7 +244,11 @@ export class CommitteeApprovalManageSection extends Component {
             <Step.Content>
               <div className="committee-action-step">
                 <div>
-                  <Step.Title>{i18next.t("EP Board review")}</Step.Title>
+                  <Step.Title>
+                    {i18next.t("{{committeeName}} review", {
+                      committeeName,
+                    })}
+                  </Step.Title>
                   <Step.Description>
                     {step2Completed ? (
                       requestLink ? (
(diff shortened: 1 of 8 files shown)


assets/js/components/record_details/CommitteeApproval.js
```diff
@@ -104,6 +104,7 @@ export class CommitteeApprovalManageSection extends Component {
     const {
       can_submit: canSubmit,
       can_create_public: canCreatePublicFlag,
+      committee_name: committeeName = null,
```
palkerecsenyi: There shouldn't be any scenario where this is not defined right?

assets/js/components/record_details/CommitteeApproval.js
```diff
@@ -228,7 +236,13 @@ export class CommitteeApprovalManageSection extends Component {
             <Step.Content>
               <div className="committee-action-step">
                 <div>
-                  <Step.Title>{i18next.t("EP Board review")}</Step.Title>
+                  <Step.Title>
+                    {committeeName
```
palkerecsenyi: I think to keep the code simpler it is safe to assume that the committee name will always be defined. I can't see a scenario where we would configure an approval flow without setting the committee name

site/cds_rdm/requests/committee_approval_state.py
```diff
@@ -15,6 +15,16 @@
 from invenio_requests.proxies import current_requests_service
 
 
+def _approval_label(community_config=None):
+    """Return the UI label for this workflow."""
+    return (community_config or {}).get("label") or "Committee approval"
```
palkerecsenyi: Same here, safe to assume we will configure a value

site/cds_rdm/requests/committee_approval_state.py
```diff
@@ -150,6 +162,7 @@ def get_committee_approval_state(record_ui, record=None):
     """
     # Read committee_approval from the parent.
     ea = _get_parent_committee_approval(record)
+    _, community_config = _get_enrolled_community(record_ui)
```
palkerecsenyi: Why are we calling this twice?

site/cds_rdm/requests/committee_approval_state.py
TahaKhan998: Removed.

site/cds_rdm/requests/committee_approval_state.py
TahaKhan998: Good catch, now computed once

assets/js/components/record_details/CommitteeApproval.js
TahaKhan998: Yes, i removed it now.

palkerecsenyi (commented):
Image

Looking very nice, thank you!! I think there are 2 other places where it would be nice to have the committee name:

1. Image
  
    EP specifically requested us to change this to "EP Approval request workflow" to make it clearer to users what the "approval" refers to.

2.  Image

    On the approval request page, the request type tag looks quite different to the tags of other request types. It would be nice for this to say "EP Approval". Currently these are set [here](https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/assets/semantic-ui/js/invenio_requests/contrib/labels/TypeLabel.js), I am not sure how easy it is to add an instance-specific override. Don't worry about it if it's too complicated :)

TahaKhan998:
> Member

image
image


Made those changes :)

assets/js/invenio_app_rdm/overridableRegistry/mapping.js
```diff
@@ -29,6 +32,17 @@ const RecordManagementContainer = (props) => (
   </>
 );
 
+const CommunitySubmissionRequestTypeLabel = () => {
+  const dataset = useContext(DatasetContext);
+  const approvalLabel = dataset?.request?.approval_label;
+
+  return (
+    <Label horizontal className="primary theme-secondary" size="small">
+      {approvalLabel || requestsI18next.t("Draft review")}
```
palkerecsenyi:
```suggestion
      {approvalLabel || requestsI18next.t("Committee review")}
```

assets/js/invenio_app_rdm/overridableRegistry/mapping.js
```diff
@@ -48,6 +62,8 @@ export const overriddenComponents = {
   "InvenioRdmRecords.RecordLandingPage.RecordManagement.NewVersionButton":
     NewVersionButton,
   "InvenioRequests.LockRequest": LockRequestComponent,
+  "RequestTypeLabel.layout.community-submission":
```
palkerecsenyi:
```suggestion
  "RequestTypeLabel.layout.committee-approval":
```
see `window.reactOverridableEnableDevMode()`:

Image

site/cds_rdm/requests/committee_approval.py
```diff
@@ -354,3 +367,19 @@ class CommitteeApprovalRequest(RDMBaseRequest):
...
+    def _create_marshmallow_schema(cls):
+        """Add the configured label when serializing existing requests."""
+        schema = super()._create_marshmallow_schema()
+        return schema.from_dict(
+            {
+                "approval_label": fields.Function(
+                    serialize=_community_submission_approval_label
+                )
+            }
+        )
```
palkerecsenyi: Why is this needed? And why is it inheriting from the community submission request? The new CDS-specific committee approval request is independent of the RDM community submission request.

TahaKhan998: You’re right, this was using the wrong request type. removed it now

palkerecsenyi (approved):
Looks great, thank you @TahaKhan998!! And thank you for all the changes :)

TahaKhan998:
> Looks great, thank you @TahaKhan998!! And thank you for all the changes :)

You're very welcome Pal. Thanks for reviewing it :)

PR #854, opened 2026-07-15
fix harvester conflict handling

### LINKED ISSUE #846: Harvester: reduce false conflicts by treating INSPIRE as source of truth (except CDS-DOI records)
Some harvester runs flag records as conflicts when INSPIRE and CDS metadata differ (for example metadata.publication_date: 2016-10-27 -> 2014 [year_mismatch]), which forces unnecessary manual curation in CDS even though INSPIRE is already curated. We should change conflict logic so INSPIRE is treated as the source of truth by default and records are updated automatically, while records with a CDS-managed DOI still go to conflict review because CDS remains authoritative in that case. True transformation failures (such as multiple DOI found) should stay hard errors, separate from conflicts.

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/846
This change reduces false harvester conflicts by treating INSPIRE as the source of truth for valid publication date differences. For regular records, the incoming INSPIRE publication date is now applied automatically instead of requiring manual conflict resolution. Records with a CDS-managed DOI remain on the existing strict conflict path because CDS is authoritative for those records. Genuine transformation failures and other update conflicts continue to be reported as errors.
Tests cover both regular records and CDS-DOI records to verify the two policies behave as expected.

### DIFF
--- a/site/cds_rdm/inspire_harvester/update/config.py
+++ b/site/cds_rdm/inspire_harvester/update/config.py
@@ -29,7 +29,7 @@ UPDATE_STRATEGY_CONFIG = {
     "metadata.contributors": CreatibutorsFieldUpdate(strict=False),
     "metadata.identifiers": IdentifiersFieldUpdate(),
     "metadata.related_identifiers": RelatedIdentifiersUpdate(),
-    "metadata.publication_date": PublicationDateUpdate(),
+    "metadata.publication_date": OverwriteFieldUpdate(),
     "metadata.subjects": ListOfDictAppendUniqueUpdate(key_field="subject"),
     "metadata.languages": ListOfDictAppendUniqueUpdate(key_field="id"),
     "metadata.description": OverwriteFieldUpdate(),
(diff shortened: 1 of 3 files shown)


site/cds_rdm/inspire_harvester/update/config.py
```diff
@@ -29,7 +29,9 @@
     "metadata.contributors": CreatibutorsFieldUpdate(strict=False),
     "metadata.identifiers": IdentifiersFieldUpdate(),
     "metadata.related_identifiers": RelatedIdentifiersUpdate(),
-    "metadata.publication_date": PublicationDateUpdate(),
+    "metadata.publication_date": PublicationDateUpdate(
```
kpsherva: wouldn't it be easier to use OverwriteFieldUpdate, instead of creating a new argument?

site/cds_rdm/inspire_harvester/update/config.py
```diff
@@ -39,3 +41,9 @@
     "custom_fields.cern:experiments": ListOfDictAppendUniqueUpdate(key_field="id"),
     # "custom_fields.cern:beams": IgnoreFieldUpdate(),
 }
+
+
+CDS_DOI_UPDATE_STRATEGY_CONFIG = {
```
kpsherva: this name confused me a little bit - I would suggest changing it to something more descriptive
```suggestion
CDS_ORIGINAL_RECORD_UPDATE_STRATEGY_CONFIG = {
```

site/cds_rdm/inspire_harvester/update/config.py
TahaKhan998: Good point, I changed this to use OverwriteFieldUpdate and removed the new argument.

site/cds_rdm/inspire_harvester/update/config.py
TahaKhan998: Agreed, I renamed it to CDS_ORIGINAL_RECORD_UPDATE_STRATEGY_CONFIG.

PR #855, opened 2026-07-15
fix harvester document type splitting

### LINKED ISSUE #849: Harvester - document type splitter feedback
Please create one PR per each section to facilitate the review process:
## Logic checks and tests

- [x] corner case: check if 2 document types mapped to the same resource type create 2 versions (should create only 1), write tests for this case (https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3235172138)
- [x] check if _PREPRINT_DOC_TYPES is used, if not, remove: https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3475979395
- [x] reorganise the filter implementation: https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3551572861

## Files
- [x] check if file without a checksum is treated as a changed file (when downloading directly from arxiv) - if yes, try to calculate the checksum of the incoming file on the fly and then compare with the existing one: https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3479834563
- [x] fix the pdf filename assumption mentioned in https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3465842579
- [x] move the functionality of file deletion to draft lifecycle https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3550315375

## Affiliations

- [x] check if the logic is correct as suggested in https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3480251244 the goal there is for each affiliation to take ROR affiliation if it exists, if not, fallback to the string

## Minor fixes and naming

- [x] https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3480281099
- [x] license headers https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3544883869
- [x] remove the env variable if not needed anymore (check if harvester runs) https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3544910530
- [x] rename method https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3550120688
- [x] rename class https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3550134606
- [x] move the import https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3550262694
- [x] add log https://github.com/CERNDocumentServer/cds-rdm/pull/732#discussion_r3550329619

TahaKhan998:
Part of #849, covering the Logic checks and tests section. This prevents duplicate versions when multiple INSPIRE document types map to the same CDS-RDM resource type, adds a regression test for that case, removes the unused _PREPRINT_DOC_TYPES constant, and reorganizes the record-matching filters into named dataclasses without changing their behavior or priority.

### DIFF
--- a/site/cds_rdm/inspire_harvester/load/matcher.py
+++ b/site/cds_rdm/inspire_harvester/load/matcher.py
@@ -7,7 +7,6 @@
 
 """Record matching module."""
 
-from collections import OrderedDict
 from dataclasses import dataclass, field
 from typing import List, Optional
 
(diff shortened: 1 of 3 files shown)


site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -66,59 +105,58 @@ def _retrieve_identifier(self, identifiers, scheme) -> Optional[str]:
...
-                # INSPIRE stores parent PID
-                "filter": [dsl.Q("term", **{"parent.id": cdsrdm_id})],
-                "value": cdsrdm_id,
-            },
-            cds_identifiers={
-                "filter": [
+        return [
+            ParentMatchFilter(
+                value=cdsrdm_id,
+                query=[dsl.Q("term", **{"parent.id": cdsrdm_id})],
```
kpsherva: I think the point was to also contain the query inside the respective filter class, so that we simplify the code here

TahaKhan998: I’ve moved each query into its respective filter class

kpsherva (approved):

PR #857, opened 2026-07-15
add(resource types): new conference note type

### LINKED ISSUE #422: migration: FASER
https://cds.cern.ch/collection/FASER
analysis codimd: https://codimd.web.cern.ch/pSoJZX9eR7yvNHhlOSh8wg
who can access/comment: https://codimd.web.cern.ch/pSoJZX9eR7yvNHhlOSh8wg#FASER-Access

## General analysis:
- Draft collections are just restricted records. They dont require specific implementation.

### EP Approval
- It's controlled with marc `9031_`
- We should create ep-approval request with the request date (when the status is waiting)
- Accept the request with approved date
- 2 records with same metadata(latest) restricted file in one record main files in another record (public record same as when you click create public record)

### FASER Papers (19):
- no comments, some records have EP Approval. 
    - Only this has comments: https://cds.cern.ch/record/2864686?ln=en
- **EP approval records has restricted draft file(file type: EPPHAPP_File)**
    - https://cds.cern.ch/record/2766207/files
    - 2 records should be created with the same(latest medata), one with only the EPPHAPP_File + EP Approval request + public record with main files. 

### FASER Conference Notes / FASER Conference Proceedings
- records are public but comments are hidden, even for super-admins
    - https://cds.cern.ch/search?ln=en&cc=FASER+Conference+Notes&sc=1&p=&action_search=Search&op1=a&m1=a&p1=&f1=
    - https://gitlab.cern.ch/cds-team/cds-legacy/-/blob/master/src/wn-cdsweb/lib/python/invenio/webcomment.py#L99-163
    - Since in rdm comments are not public this should be ok?
        - It will be visible to submitter, reviewer, community curators etc. Is it fine?

### TODO
- [ ] check if 1 record belongs 2 communities and there's comments for one, does the other one can access?
- [X] check if there is any ep approval UI interface to comment? or it's offline comments?
    - checked there's no ep approval comments. it's only email and marc tag
- [ ] check access: one community shouldnt see other's comments?
- [ ] Multiple descriptions, we should handle this: https://cds.cern.ch/record/2936325 
    - will go to additional desc
- [x] Add this affiliation: https://ror.org/02avf8f85 record: https://cds.cern.ch/record/2816452

### Library
- Do they want to be able to curate only cern-ep public records or other public records from faser?

### FASER Community
- Most of the FASER records(the ones without EP workflow) has 591__:approved. Is this something needed? Since these records will have community submission request(if they have comments)? Is it important to keep when it's approved? Can we just migrate the final record?
    - example: https://cds.cern.ch/record/2853210?ln=en 
- For the EP report number records, since we need to create 2 records to keep restricted draft file, can we use the same(latest) metadata in both records? Or is it important to use the metadata before it's approved for draft records. (We can use the latest revision before approved date)
- In legacy cds there's no commenting for EP approval. If there's EP approval + comments it'll be the comments of community submission/or they want the comments in EP aproval request?
- After EP approval, submitter can't edit any metadata of the record. only super-editor can edit.
    - https://gitlab.cern.ch/cds-team/cds-legacy/-/blob/master/src/wn-cdsweb/lib/python/invenio/websubmit_functions/GENSBM_utils.py#L1039-1094
    - Is it okay of it's allowed to edit? Or the new created public record(like in the demo we dont keep the submitter as submitter?)
        - what is the case?
- FASER Conference Notes / FASER Conference Proceedings comments are hidden even for super-admins
    - https://cds.cern.ch/search?ln=en&cc=FASER+Conference+Notes&sc=1&p=&action_search=Search&op1=a&m1=a&p1=&f1=
    - https://gitlab.cern.ch/cds-team/cds-legacy/-/blob/master/src/wn-cdsweb/lib/python/invenio/webcomment.py#L99-163
    - Since in rdm comments are not public this should be ok?
        - It will be visible to submitter, reviewer, community curators etc. Is it fine?

### Migration About EP Records
- Since we have same metadata for 2 records (restricted(the one requested the ep approval) + public record(the one goes to CERN Research)) we should mint legacy recid(effects redirection) report number doi etc. Which one we should mint?
    - Redirections to which record? Public one?
    - When we save original dump(marcxml) which record we use? we need migrated parent record uid
    - Stats migration, which record? both?
    - CLC sync, is this the case?

zubeydecivelek:
closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/422

### DIFF
--- a/app_data/vocabularies/resource_types.yaml
+++ b/app_data/vocabularies/resource_types.yaml
@@ -86,6 +86,28 @@
   tags:
     - depositable
     - linkable
+- id: publication-conferencenote
+  icon: file alternate
+  props:
+    csl: report
+    datacite_general: Text
+    datacite_type: Conference note
+    openaire_resourceType: "0020"
+    openaire_type: publication
+    eurepo: info:eu-repo/semantics/other
+    schema.org: https://schema.org/CreativeWork
+    subtype: publication-conferencenote
+    type: publication
+    marc21_type: publication
+    marc21_subtype: conferencenote
+  title:
+    en: Conference note
+    sv: Konferensanteckning
+    de: Konferenznotiz
+    cs: Konferenční poznámka
+  tags:
+    - depositable
+    - linkable
 - id: publication-conferenceproceeding
   icon: file alternate
   props:


app_data/vocabularies/resource_types.yaml
```diff
@@ -86,6 +86,28 @@
   tags:
     - depositable
     - linkable
+- id: publication-conferencenote
+  icon: file alternate
+  props:
+    csl: paper-conference
```
kpsherva: I think the other integrations should not map to conference paper because they are not always being published  on the conference, like Jamie mentioned

zubeydecivelek: updated! is it correct now?

kpsherva (approved):

PR #866, opened 2026-07-20
fix(harvester): normalize malformed URN identifiers

### LINKED ISSUE #861: thesis harvester: verification

Image

Image

TahaKhan998:
Part of #861.
Solves no.7 of https://codimd.web.cern.ch/s/DBWeY3d5y

This fixes the harvester failure metadata.related_identifiers.identifier: Invalid URN identifier, which was found for INSPIRE#1714668. INSPIRE marks the identifier as a URN but provides it in a malformed form such as http://nbnurn:nbn:de:..., while CDS-RDM expects it to start with urn:. The related identifier mapper now extracts the actual urn: value, validates and normalizes it with idutils, and only adds it when it is valid. Values that cannot be safely corrected are skipped and reported as transformation errors instead of reaching CDS record validation. The same handling is used for both persistent and external identifiers, while the existing behavior for valid URNs, DOIs, and other identifier schemes stays unchanged. Tests cover the real failing value, existing valid URNs, and values that cannot be corrected.

image
image

### DIFF
--- a/site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
+++ b/site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
@@ -11,10 +11,53 @@ import json
 from dataclasses import dataclass
 
 from flask import current_app
-from idutils.normalizers import normalize_isbn
-from idutils.validators import is_doi
+from idutils.normalizers import normalize_isbn, normalize_urn
+from idutils.validators import is_doi, is_urn
 
 from cds_rdm.inspire_harvester.transform.mappers.mapper import MapperBase
+
+
+def _coerce_urn(value):
+    """Return a CDS-valid URN, or ``None`` if it cannot be safely corrected.
+
+    INSPIRE sometimes stores URNs with an ``http(s)://`` prefix, e.g.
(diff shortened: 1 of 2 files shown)


sakshamarora1 (approved):

site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
```diff
@@ -11,12 +11,55 @@
...
+            ctx.errors.append(f"Invalid URN. | details: value={original}")
+            return None
+
+    related = {
+        "identifier": value,
+        "scheme": schema,
+        "relation_type": {"id": "isvariantformof"},
+        "resource_type": {"id": "publication-other"},
+    }
+    if schema == "doi":
```
kpsherva, line 58: I think with this approach we will always map DOI to related identifiers, could you verify ithe implementation: if we have CDS DOI or only one DOI it is mapped to "pids"?

TahaKhan998: Good catch but I double checked and this doesn't change the DOI behaviour. The if schema == "doi" part in _related_identifier is the exact same code that was already inline in both loops on master, I just moved it into the helper so the URN handling has one place to live.

kpsherva (approved):

PR #869, opened 2026-07-21
fix(harvester): merge duplicate contributors

### LINKED ISSUE #861: thesis harvester: verification

TahaKhan998:
Closes #861

The harvester was failing when a CDS record contained multiple contributor entries matching the same incoming INSPIRE contributor. Since it could not decide which entry to update, it raised a `Multiple creators match incoming` conflict.

This PR merges the matching CDS contributor entries before applying the incoming INSPIRE data. Metadata, identifiers, and affiliations from the existing entries are preserved while exact duplicates are removed.

It also fixes the `duplicated affiliations` errors that appeared when the same records were harvested again. The affiliation merge previously checked duplicate names but always appended ID-based affiliations, so repeated runs could add the same vocabulary ID more than once. Affiliations are now deduplicated by both ID and name across the stored and incoming data, including duplicates already present in the CDS record.

Finally, strict update conflicts are logged only by the writer instead of both the update engine and writer, so the same error no longer appears twice in the harvester report.image

image

### DIFF
--- a/site/cds_rdm/inspire_harvester/update/fields/creatibutors.py
+++ b/site/cds_rdm/inspire_harvester/update/fields/creatibutors.py
@@ -33,7 +33,7 @@ class CreatibutorsFieldUpdate(FieldUpdateBase):
         self.strict = strict
 
     def _union_affiliations(self, cur_list, inc_list):
-        """Union affiliations by name.
+        """Union affiliations by identifier or name.
 
         Exact duplicates are skipped. If an incoming affiliation differs from an
         existing one only in punctuation, the existing entry is updated in-place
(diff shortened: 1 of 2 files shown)


site/cds_rdm/inspire_harvester/update/fields/creatibutors.py
```diff
@@ -92,10 +94,57 @@ def _key(self, creator: dict):
...
+                continue
+            first_value = first.get(key)
+            second_value = second.get(key)
+            if (
+                first_value not in empty_values
+                and second_value not in empty_values
+                and first_value != second_value
+            ):
+                return True
+        return False
```
sakshamarora1: let's write this in a more readable way with using proper naming for the variables as well. it'll help later in understanding the flow / debugging

TahaKhan998: pushed a simpler version with clearer variable names (first_value, second_value, etc.). Happy to tweak further if anything still reads awkwardly.

site/cds_rdm/inspire_harvester/update/fields/creatibutors.py
```diff
@@ -92,10 +94,45 @@ def _key(self, creator: dict):
...
+            if first_value != second_value:
+                return True
+        return False
+
+    def _duplicates_are_compatible(self, creators):
+        """Check that every matching creator can be safely merged into one."""
+        for i, first in enumerate(creators):
+            for second in creators[i + 1 :]:
+                if self._have_conflicting_values(
+                    first, second, ignored=("affiliations", "person_or_org")
```
kpsherva, line 115: I think the conflicting affiliation should not be ignored - it might be that a person changed a job during the creation of a record or the affiliation was wrongly registered in the first place and should be checked by a curator

kpsherva:
I need a walk-through the code from the creator to understand what is happening :)

TahaKhan998:
Affected records should be fixed by curators

PR #873, opened 2026-07-22
Ep lep workflow

kpsherva:


invenio.cfg
```diff
@@ -459,20 +454,73 @@ CDS_CERN_SCIENTIFIC_COMMUNITY_ID = ""
...
-    #     "report_number": {
-    #         "prefix": "CERN-EP",   # literal prefix, e.g. "CERN-EP"
-    #         "include_year": True,  # append the current year after prefix
-    #         "counter_digits": 3,   # zero-padding width, e.g. 3 → "001"
-    #     },
-    # },
+    "dd13404c-bcd6-4b15-aeef-38d678c61ff1": {
+        # faser
+        "label": "EP approval",  # shown in UI buttons/headings
+        "referee_group": "cds-ph-ep-publication",  # CERN e-group slug
```
palkerecsenyi, line 461: Here we need to add the `committee_name`. I guess for all of them it will be the same:


```suggestion
        "referee_group": "cds-ph-ep-publication",  # CERN e-group slug
        "committee_name": "EP Board",
```

kpsherva: in the case of the migrator is not relevant - here we need just the prefix, but when we actually set it up on the cds-rdm then it will matter

palkerecsenyi: Yes indeed, on dev/sandbox/prod we will use the configmap to set this. But I'm just mentioning it here since it would be useful for us to have the correct labels on our local instances in case we're testing it.

PR #874, opened 2026-07-22
feat(ep): allow referees to view new record versions

### LINKED ISSUE #864: Ensure EP referees can see newer record versions
EP referees should be able to see not only the version of the restricted record for which the approval request was made, but also all newer versions of the record. I don't know if this is the case right now, we should check and fix if not.

palkerecsenyi:
Closes #864 

---

* Changed the permissions so that EP referees are able to view versions of the restricted/internal record that were created after the approval was requested.

* Referees can now access:
  - The record submitted for review
  - All future versions of that record

### DIFF
--- a/site/cds_rdm/requests/committee_approval.py
+++ b/site/cds_rdm/requests/committee_approval.py
@@ -67,7 +67,9 @@ def _resolve_community_config(request):
             "The record is not part of any community. "
             "It must belong to a committee-approval-enabled community before submitting."
         )
-    committee_communities = current_app.config.get("CDS_COMMITTEE_APPROVAL_COMMUNITIES", {})
+    committee_communities = current_app.config.get(
+        "CDS_COMMITTEE_APPROVAL_COMMUNITIES", {}
+    )
     if default_community_id in committee_communities:
         return committee_communities[default_community_id]
     raise ValidationError(
@@ -98,8 +100,10 @@ class CommitteeApprovalSubmitAction(actions.CreateAndSubmitAction):
         config = _resolve_community_config(self.request)
 
         # Reject if the parent already carries an approval report number.
-        if ((topic.parent.get("permission_flags") or {}).get("committee_approval") or {}).get(
-            "reportnumber"
+        if (
+            topic.parent.get("permission_flags", {})
+            .get("committee_approval", {})
+            .get("reportnumber")
         ):
             raise ValidationError(
                 "A version of this record already has an approval report number assigned. "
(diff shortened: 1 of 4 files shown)


site/cds_rdm/requests/committee_approval.py
```diff
@@ -146,7 +148,7 @@ def execute(self, identity: Identity, uow: UnitOfWork) -> None:
             subject_type="role",
             subject_id=config["referee_group"],
             permission=COMMITTEE_APPROVAL_GRANT_PERMISSION,
-            origin=committee_approval_grant_origin(topic.id),
+            origin=committee_approval_grant_origin(topic.versions.index),
```
zzacharo: I would rather not remove the topic.id from the from grant as it is useful to identify the record that it refers to. The version.index is quite weak in that sense. Mayby we can concatenate it? `topic_id-version_index`?

site/tests/test_committee_approval.py
```diff
@@ -540,11 +547,9 @@ def test_referee_grant_added_on_submit_removed_on_decline(
 
     request_type = current_request_type_registry.lookup("committee-approval")
 
-    pid_obj = PersistentIdentifier.get(
-        "recid", record_in_enrolled_community.id
-    )
+    pid_obj = PersistentIdentifier.get("recid", record_in_enrolled_community.id)
     record_uuid = pid_obj.object_uuid
-    expected_origin = f"{COMMITTEE_APPROVAL_GRANT_ORIGIN_PREFIX}{record_uuid}"
+    expected_origin = f"{COMMITTEE_APPROVAL_GRANT_ORIGIN_PREFIX}{record_uuid}_1"
```
zzacharo: can we instead use the pid value? It would be easier to track down I feel.... @kpsherva thoughts? This is the grant key stored in the parent of giving access to referees. I am also happy to merge like this if you think I am overthinking it.

zzacharo (approved):

site/tests/test_committee_approval.py
kpsherva: I need a reminder why we need to do this in a first place to answer this

zzacharo: we give a grant access to a role so that the referree group can see the restricted record that they review

palkerecsenyi: We could change it to the PID value, I don't think it would make a huge difference. But i'm not sure it will help us much either except in some very niche circumstances.

PR #875, opened 2026-07-23
fix(custom_fields): serialize legacy meeting objects as a list

### LINKED ISSUE #865: UI backward compatibility for object conference field
In #844 we changed the conference field to be an array instead of an object, thereby allowing for multiple conferences per record. We need to update the UI handlers (view handlers, templates, etc wherever else it is affected) to be backward-compatible with existing records that use objects for the conference field. We don't want to update each record in the DB/OS.

TahaKhan998:
Closes #865
This pr adds a serialize-only compatibility list field that wraps a legacy object as a one-item list for reads, leaves arrays unchanged.

### DIFF
--- a/site/cds_rdm/custom_fields/meeting.py
+++ b/site/cds_rdm/custom_fields/meeting.py
@@ -49,34 +49,53 @@ MeetingDublinCoreDumper.post_dump = _support_meeting_list(
 )
 
 
+class BackwardCompatibleMeetingList(fields.List):
+    """Serialize a legacy meeting object as a one-item list.
+
+    Older records store ``meeting:meeting`` as a single object. After the
+    multi-conference change it is an array. Only serialization is adapted so
+    reads present a list to the UI; writes must still submit an array.
+    """
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        """Wrap a legacy dict in a list without mutating storage."""
+        if isinstance(value, dict):
+            value = [value]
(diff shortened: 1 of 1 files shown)


site/cds_rdm/custom_fields/meeting.py
```diff
@@ -49,34 +49,53 @@ def wrapper(self, data, original=None, **kwargs):
 )
 
 
+class BackwardCompatibleMeetingList(fields.List):
```
kpsherva, line 52: Thanks for taking this one! wouldn't it be easier to implement this only on the new jinja template (CDSMeeting)? Did you consider that?

TahaKhan998: I did think about that. A Jinja only fix in CDSMeeting.html would cover the record page, but the edit form still loads custom_fields through the schema dump into ArrayField, so a legacy object would still break there. Doing the wrap in serialize felt cleaner so both paths get a list without touching the widget or allowing object writes.

palkerecsenyi: Yeah also the API response serializes it incorrectly too if only the jinja template is changed

palkerecsenyi (approved):
perfect thanks @TahaKhan998! I think this is a very nice elegant simple way to do it, I can't see any problems with it. We will make sure to test in dev with object meeting field records (we might need to create some manually).

PR #877, opened 2026-07-23
fix(harvester): put CERN- report numbers in identifiers

### LINKED ISSUE #861: thesis harvester: verification

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/861
INSPIRE report numbers starting with CERN- (scheme cdsrn) were always mapped into metadata.related_identifiers but they now go to metadata.identifiers, while non-CERN- report numbers stay related. The record matcher reads and searches cdsrn in identifiers to match that placement.

### DIFF
--- a/site/cds_rdm/inspire_harvester/load/matcher.py
+++ b/site/cds_rdm/inspire_harvester/load/matcher.py
@@ -104,16 +104,52 @@ class ArxivIdentifierMatchFilter(FilterCandidate):
 
 @dataclass(frozen=True)
 class ReportNumberMatchFilter(FilterCandidate):
-    """Match a CDS report number."""
+    """Match a CDS report number in metadata.identifiers."""
 
     @property
     def query(self):
         """Build the CDS report number query."""
         return [
+            dsl.Q("term", **{"metadata.identifiers.scheme": "cdsrn"}),
+            dsl.Q(
+                "term",
+                **{"metadata.identifiers.identifier": self.value},
+            ),
+        ]
+
+
+@dataclass(frozen=True)
+class RelatedReportNumberMatchFilter(FilterCandidate):
+    """Match a CDS report number in metadata.related_identifiers."""
+
+    @property
+    def query(self):
+        """Build the related CDS report number query."""
+        return [
             dsl.Q("term", **{"metadata.related_identifiers.scheme": "cdsrn"}),
             dsl.Q(
                 "term",
                 **{"metadata.related_identifiers.identifier": self.value},
+            ),
+            dsl.Q(
+                "term",
+                **{"metadata.related_identifiers.relation_type.id": "isvariantformof"},
+            ),
+        ]
+
+
+@dataclass(frozen=True)
+class ApprovalReportNumberMatchFilter(FilterCandidate):
+    """Match an EP/approval report number (apprn) in metadata.identifiers."""
+
+    @property
+    def query(self):
+        """Build the approval report number query."""
+        return [
+            dsl.Q("term", **{"metadata.identifiers.scheme": "apprn"}),
+            dsl.Q(
+                "term",
+                **{"metadata.identifiers.identifier": self.value},
             ),
         ]
 
@@ -163,11 +199,14 @@ class RecordMatcher:
     def _build_filter_priority(self, entry, inspire_id, cdsrdm_id):
         """Build the priority-ordered record match candidates."""
         doi = entry.get("pids", {}).get("doi", {}).get("identifier")
+        identifiers = entry["metadata"].get("identifiers", [])
         related_identifiers = entry["metadata"].get("related_identifiers", [])
 
         cds_id = self._retrieve_identifier(related_identifiers, "cds")
         arxiv_id = self._retrieve_identifier(related_identifiers, "arxiv")
-        report_number = self._retrieve_identifier(related_identifiers, "cdsrn")
+        report_number = self._retrieve_identifier(identifiers, "cdsrn")
+        related_report_number = self._retrieve_identifier(related_identifiers, "cdsrn")
+        approval_report_number = self._retrieve_identifier(identifiers, "apprn")
         return [
             ParentMatchFilter(value=cdsrdm_id),
             CDSIdentifierMatchFilter(value=cds_id),
@@ -175,4 +214,6 @@ class RecordMatcher:
             InspireIdentifierMatchFilter(value=inspire_id),
             ArxivIdentifierMatchFilter(value=arxiv_id),
             ReportNumberMatchFilter(value=report_number),
+            ApprovalReportNumberMatchFilter(value=approval_report_number),
+            RelatedReportNumberMatchFilter(value=related_report_number),
         ]
--- a/site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
+++ b/site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
@@ -8,12 +8,14 @@
 """INSPIRE to CDS harvester module."""
 
 import json
+import re
 from dataclasses import dataclass
 
 from flask import current_app
 from idutils.normalizers import normalize_isbn, normalize_urn
 from idutils.validators import is_doi, is_urn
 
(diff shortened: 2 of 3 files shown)


site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
```diff
@@ -211,10 +218,13 @@ def map_value(self, src_record, ctx, logger):
 
             report_numbers = src_metadata.get("report_numbers", [])
             for rn in report_numbers:
+                value = rn.get("value")
+                if not value or value.startswith("CERN-"):
+                    continue
                 identifiers.append(
```
sakshamarora1, line 297: can we name this properly to avoid confusion please?

site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
```diff
@@ -211,10 +218,13 @@ def map_value(self, src_record, ctx, logger):
...
                     {
                         "scheme": "cdsrn",
```
sakshamarora1, line 299: will non `CERN-` identifiers still be cdsrn?

site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
TahaKhan998: Renamed to `report_number`.

site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
TahaKhan998: Yes, non-`CERN-` report numbers stay as related identifiers with scheme `cdsrn` (same as before). Only placement of `CERN-` ones changed.

site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
```diff
@@ -209,12 +216,16 @@ def map_value(self, src_record, ctx, logger):
                     }
                 )
 
+            # Non-CERN- report numbers stay related identifiers (still scheme cdsrn).
             report_numbers = src_metadata.get("report_numbers", [])
             for rn in report_numbers:
+                report_number = rn.get("value")
+                if not report_number or report_number.startswith("CERN-"):
```
kpsherva: also, we need to monitor for EP numbers now... and the schema for EP numbers is different. 
it should be taken into account when searching for existing records too - we didn't use the apprn until now, could you add matching by the EP approval too?

TahaKhan998: Added. EP numbers now map to apprn (using the prefixes from CDS_COMMITTEE_APPROVAL_COMMUNITIES), and the matcher also searches by apprn when looking for existing records. Other CERN- report numbers still go to identifiers as cdsrn.

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -163,16 +195,21 @@ def _retrieve_identifier(self, identifiers, scheme) -> Optional[str]:
...
         doi = entry.get("pids", {}).get("doi", {}).get("identifier")
+        identifiers = entry["metadata"].get("identifiers", [])
         related_identifiers = entry["metadata"].get("related_identifiers", [])
 
         cds_id = self._retrieve_identifier(related_identifiers, "cds")
         arxiv_id = self._retrieve_identifier(related_identifiers, "arxiv")
-        report_number = self._retrieve_identifier(related_identifiers, "cdsrn")
+        report_number = self._retrieve_identifier(identifiers, "cdsrn")
+        related_report_number = self._retrieve_identifier(related_identifiers, "cdsrn")
+        approval_report_number = self._retrieve_identifier(identifiers, "apprn")
```
kpsherva, line 209: @zzacharo @palkerecsenyi guys, how are we serializing approval report number when sending it to datacite? in general, in export formats we should not have another identifier type for apprn, it should be indistinguishable from reportnumber, since apprn is our internal system implementation

kpsherva (approved):

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -163,16 +195,21 @@ def _retrieve_identifier(self, identifiers, scheme) -> Optional[str]:
...
         return [
             ParentMatchFilter(value=cdsrdm_id),
             CDSIdentifierMatchFilter(value=cds_id),
             DOIMatchFilter(value=doi),
             InspireIdentifierMatchFilter(value=inspire_id),
             ArxivIdentifierMatchFilter(value=arxiv_id),
             ReportNumberMatchFilter(value=report_number),
+            RelatedReportNumberMatchFilter(value=related_report_number),
+            ApprovalReportNumberMatchFilter(value=approval_report_number),
```
kpsherva: I would set approval report number as a priority since it is more "controlled" by the system
Matching by related report number might be very error-prone - you will match records which are related to the record you are actually searching for and it might return multiple matches quite frequently. I think you might add some more restrictive criteria to this. Can you propose what would make it less error prone?

TahaKhan998: the approach i used is to make the db search stricter by searching related identifiers with relation_type isvariantformof and scheme cdsrn, and if it still returns multiple records then treat it as ambiguous. also put apprn above related report number in the priority list

site/cds_rdm/inspire_harvester/load/matcher.py
palkerecsenyi: It is indistinguishable if I understand correctly. For example, https://dev-cds-rdm.web.cern.ch/records/vq903-1p243 is exported as:

```json
  "alternateIdentifiers": [
    {
      "alternateIdentifier": "https://dev-cds-rdm.web.cern.ch/records/vq903-1p243",
      "alternateIdentifierType": "URL"
    },
    {
      "alternateIdentifier": "oai:127.0.0.1:5000:vq903-1p243",
      "alternateIdentifierType": "oai"
    },
    {
      "alternateIdentifier": "CERN-EP-2024-309",
      "alternateIdentifierType": "CDS"
    },
    {
      "alternateIdentifier": "2917427",
      "alternateIdentifierType": "CDS"
    }
  ],
```

All CDS internal identifiers are exported as type `CDS`: https://github.com/CERNDocumentServer/cds-rdm/blob/ef723ca91eef50c85803913e243bb84d1cf0debb/invenio.cfg#L696

palkerecsenyi (approved):
Looks good to me, the approval number validation is correct as far as I can tell

site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
```diff
@@ -60,6 +61,30 @@ def _related_identifier(schema, value, ctx):
...
+
+    Requires both a configured committee prefix and a value accepted by the
+    ``apprn`` scheme validator, so prefix look-alikes fall back to ``cdsrn``
+    instead of failing record validation.
+    """
+    if not value:
+        return False
+    if not schemes.is_approval_report_number(value):
+        return False
+    return any(value.startswith(prefix) for prefix in _committee_approval_prefixes())
```
zubeydecivelek: I'm not sure startswith is the best way to detect approval numbers since we might have `CERN-EP-DRAFT-....` report numbers

palkerecsenyi: Hmmmm this is true, I am not sure what would be the best way to solve it though. We could specifically check for `CERN-EP-DRAFT-*` and exclude it, not sure if there is any other reliable way of doing it

zubeydecivelek: can we do a regex and check if the report number is in this format `prefix-digit*` so if we have `prefix-letter` we exclude?

TahaKhan998: yup good catch, i have added the regex you suggested

zubeydecivelek (approved):

PR #884, opened 2026-07-27
fix(harvester): error on editions and map multi DOIs

### LINKED ISSUE #501: [INSPIRE Harvester] Open questions around mapping
## 1. Authors Affiliations (`authors.affiliations` / `record_affiliations`)

**Current fields:**
- `authors.affiliations.value`
- `record_affiliations.value`

**Other identifiers:**
- `authors.affiliations_identifiers.schema`
- `authors.affiliations_identifiers.value`

**Questions:**
- What is the plan for exposing or storing the ROR identifier?
- Should we store multiple affiliation identifiers?

## 2. Editions Field

In Inspire, theses typically do not have an `editions` field, but other record types might.

**Potential mapping options:**
1. Add to `imprint`
2. Use `version` field
3. Add to `description` with type `other`
4. Add to `title`

**Question:**
- Which approach is btter for mapping edition information?

## 3. Funding Information (`funding_info`)

Inspire includes detailed funding information.

**Questions:**
- Should we import extensive funding data? (f.e. matching the vocabularies used in Zenodo)
- Is it important for researchers to have all funding/grant info imported into CDS?

## 4. Number of Pages

Inspire provides `number_of_pages`.

In CDS-RDM, the preview feature already shows the number of pages for uploaded files.

**Question:**
- Can we safely drop this field or is there a use case for keeping it?

## 5. Publication Type

Inspire provides `publication_type`.

**Questions:**
- How is this field currently used in INSPIRE?

## 6. Texkeys

Inspire includes `texkeys`.

**Questions:**
- How is this being used?

## 7. Withdrawn Records

Inspire provides a `withdrawn` flag.

**Questions:**
- Should we restrict access to files if `withdrawn = true`?
- Any other special handling for withdrawn records?

## 8. DOIs

Inspire records may contain multiple DOIs.

**Observations:**
- If a DOI has a CERN prefix, we could use it as the main DOI.
- Other DOIs could be added as related identifiers or alternative identifiers.

**Questions:**
- Should CERN DOIs always be prioritized as the primary?
- If no CERN DOI exists, how should we select the main DOI?
- Should we import all other DOIs as related/alternative identifiers?

TahaKhan998:
This PR does the editions and DOI parts of #501.  if INSPIRE sends `metadata.editions`, we raise a curator error and do not map it to `imprint:imprint.edition`. (ForCDS currently has 0 records with editions).
For DOIs, if there’s a CDS/DataCite-prefix DOI we use that as the main PID and put the others into related identifiers (isversionof). If there’s no CDS DOI and more than one other DOI, we still raise an error like before so curators can handle it. Also added tests.

in the writer: when publishing a new version we still clear pids.oai and non-external pids.doi so fresh ones can be minted, but we do it with safe pops instead of del. Same behaviour as before, it just doesn’t crash when those keys aren’t present (which happens more often now that transformed entries often only carry a DOI).

### DIFF
--- a/site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
+++ b/site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
@@ -28,7 +28,12 @@ class DOIMapper(MapperBase):
         return True
 
     def map_value(self, src_record, ctx, logger):
-        """Mapping of record dois."""
+        """Mapping of record dois.
+
+        Prefer a CDS/DataCite-prefix DOI as the main PID. Any other DOIs become
+        related identifiers (via ctx.extra_related_dois). Multiple non-CDS DOIs
+        with no CDS DOI remain an error.
+        """
         src_metadata = src_record.get("metadata", {})
         DATACITE_PREFIX = current_app.config["DATACITE_PREFIX"]
         dois = src_metadata.get("dois", [])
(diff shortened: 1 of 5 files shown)


site/cds_rdm/inspire_harvester/transform/mappers/identifiers.py
```diff
@@ -47,25 +52,55 @@ def map_value(self, src_record, ctx, logger):
...
+
+        if len(cds_dois) > 1:
+            ctx.errors.append(
+                "More than 1 CDS DOI was found. "
+                f"| details: dois={[d.get('value') for d in cds_dois]}"
+            )
             return None
-        elif len(unique_dois) == 0:
+
+        if cds_dois:
```
kpsherva, line 74: one more case - if there is only 1 external DOI, it should be mapped to "pids" as external - is this covered?

TahaKhan998: Yeah that case is covered. If there’s no CDS DOI and only one external DOI, it goes through the elif len(other_dois) == 1 branch and lands in pids with provider: external. There’s also a test for it (test_transform_dois_valid_external).

PR #886, opened 2026-07-29
fix(harvester): recover records skipped by live pagination

### LINKED ISSUE #437: hervester: make sure all records were processed
Harvester goes from page to page based on a search query in inspire.
It means while we are paginating, someone might update one of the records in the list, lets say
we have records 
```
page1 = {1,2,3}
page2 = {4,5,6}
total = 6
```
if record 3 gets updated while the harvester is still processing record 2
before we get to page2, the pages will change as follows:
```
page1={1,2,4}
page2={5,6}
```
meaning record 4 will never be picked up, harvester will not know to go back to the previous page to re-read the record 4 from there

because of this, we need to check if all the record from initial `hits.total` were processed.
if not, we need to re-run the harvest for the missing record. How do we know which one is missing?

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/437
INSPIRE search results can change while we're paginating. If a record gets updated mid-harvest, it can jump to a page we already passed, so we never pick it up even though hits.total said it was there.
What we do now: during the normal harvest we keep the IDs we actually saw, plus the total from the first page. If those don't match at the end, we re-query the same search with fields=id to get the full ID list, figure out what's missing (all_ids - seen_ids), and fetch only those records with q=... AND id:... like we already do for single-id harvests. If everything matched on the first pass, we don't do any of that extra work.


site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -40,41 +42,112 @@ def __init__(
...
+                    f"Sending INSPIRE record #{record_id} to transformer."
+                )
+                yield inspire_record
 
             # Get the next page URL if available
             url = data.get("links", {}).get("next")
 
+        # Pagination can skip records when INSPIRE results shift mid-harvest.
+        # A total mismatch can also mean records entered/left the live query.
+        if not q or expected_total is None or len(seen_ids) == expected_total:
```
kpsherva: can a record be added to the set and cause a shift in another direction? what if one record is added and another removed?

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -40,41 +42,112 @@ def __init__(
...
+
+        current_app.logger.warning(
+            "Processed fewer INSPIRE records than expected. "
+            f"| details: processed={len(seen_ids)}, expected={expected_total}"
+        )
+
+        # Re-scan with IDs only, then harvest the missing records by id.
+        all_ids = set()
+        ids_url = self._build_url(q, fields="id")
+        while ids_url:
```
kpsherva: what if something shifts in the rescan?

site/cds_rdm/inspire_harvester/reader.py
TahaKhan998: yeah good point, if one record drops out and another joins the total can stay the same and we'd never hit recovery with the count check alone.

i changed it to always do the id scan at the end and recover whatever's in all_ids - seen_ids, so we're comparing the actual sets not just the count.

site/cds_rdm/inspire_harvester/reader.py
TahaKhan998: Yes the rescan can shift too since it's still paginated.

i changed so now we run the id scan twice with size=1000, and if the two passes don't match we union them and log a warning. anything still missing gets fetched individually by id instead of re-paginating the full harvest. i also added a test for this case(test_reader_recovers_when_count_matches).

i also have another approach in mind where we collect all ids upfront at the start and then harvest each record by id instead of paginating through search at all. that'd be more robust but more expensive api-wise, so i kept this for now since it fits the current flow. happy to switch if you'd prefer that.

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -40,40 +42,134 @@ def __init__(
...
+
+        # One ID-only scan, then harvest anything we missed.
+        # Retry only if the scan collected fewer IDs than INSPIRE reported.
+        all_ids, scan_total = self._scan_ids(q, headers)
+        if scan_total is not None and len(all_ids) != scan_total:
+            current_app.logger.warning(
+                "ID scan found fewer INSPIRE records than reported. "
+                f"| details: found={len(all_ids)}, reported={scan_total}"
+            )
+            retry_ids, _ = self._scan_ids(q, headers)
```
palkerecsenyi: Why would the initial scan for IDs return a total hits count that's less than the number of hits it returned? Surely within one single API request the count would be consistent? Maybe I am misunderstanding this

palkerecsenyi (approved):
Looks good, it's a nice and efficient way of doing it! I just have one question

site/cds_rdm/inspire_harvester/reader.py
TahaKhan998: It’s not one request, _scan_ids paginates. When we first request records from a query, the first page says we have X records in this query, and that’s what scan_total is. Then when we go through [each record and each page](https://github.com/CERNDocumentServer/cds-rdm/pull/886/changes/BASE..2f637d7e60d90fcc86b9e6886680126632f8d252#diff-702683ff38e3e6b43b77940275c823555266d0b1bd9f53757445d0049b8ff542R75-R77), shifting might happen while that happens and we might miss a few records while the ids are being scanned. Then scan_total and the number of ids might not be the same, so we refetch all ids with retry_ids and union the differences between all_ids and retry_ids.

palkerecsenyi: Ahh okay makes sense, thanks for the explanation.

kpsherva: the scan IDs will have the same problem as if we iterate record by record, I would not perform two big paginated requests since they will have the same issue... especially if we will have to deal with big harvests.
isn't it enough to fetch the total number from the first page and then count each processed record, and comparing the numbers? or this is about knowing which records to retry?

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -40,40 +42,134 @@ def __init__(
...
+            data = self._get_json(url, headers)
+            if reported_total is None:
+                reported_total = data["hits"]["total"]
+            for hit in data["hits"]["hits"]:
+                ids.add(str(hit["id"]))
+            url = data.get("links", {}).get("next")
+
+        return ids, reported_total
+
+    def _iter(self, url, q=None, *args, **kwargs):
```
kpsherva: shouldn't q be a required argument? what values are passed there? I am asking because that was not the part of the previous signature of the function

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -40,40 +42,134 @@ def __init__(
...
+            if url:
+                had_another_page = True
+
+        # If results moved while we were paging, we may have missed some records.
+        # Skip this check for single-id jobs, or when everything fit on one page.
+        if (
+            not q
+            or expected_total is None
+            or self._inspire_id
+            or not had_another_page
```
kpsherva: why do we quit the loop on these conditions?

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -40,40 +42,134 @@ def __init__(
...
+        ):
+            return
+
+        # One ID-only scan, then harvest anything we missed.
+        # Retry only if the scan collected fewer IDs than INSPIRE reported.
+        all_ids, scan_total = self._scan_ids(q, headers)
```
kpsherva: I realised I don't understand the order of actions that are fired in this workflow. Could we meet to discuss? I need you to walk me through what is happening :)

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -40,40 +42,81 @@ def __init__(
...
-        while url:  # Continue until there is no "next" link
-            current_app.logger.info(f"Querying URL: {url}.")
-            response = requests.get(url, headers=headers)
-            data = response.json()
-            if response.status_code == 200:
-                current_app.logger.debug("Request response is successful (200).")
+        harvest_url = url
+        seen_ids = set()
+        first_pass = True
+
```
palkerecsenyi, line 82: maybe write a small comment explaining briefly what this does

palkerecsenyi (approved):

site/cds_rdm/inspire_harvester/reader.py
TahaKhan998: done

sakshamarora1 (approved):

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -40,40 +42,89 @@ def __init__(
...
+        # - middle (page_url): walk INSPIRE pagination (next page links)
+        # - inner (hits): yield each record on the current page
+        # seen_ids tracks ids already yielded in this run so retries do not send
+        # the same record twice. first_pass makes us retry at least once. if a
+        # later pass adds nothing new (new_in_pass == 0), harvesting stops.
+        harvest_url = url
+        seen_ids = set()
+        first_pass = True
+
+        while True:
```
kpsherva: how are you making sure this doesn't run in forever loop? Does it make sense to cap the possible repetitions to 3 times?

TahaKhan998: My idea was that it would stop once it detects that no new record has been harvested in the current pass, which is what the variable new_in_pass is for, but yes it makes sense to stop after 3 passes if the records keep shifting. I now capped the re-harvest loop to 3 passes so it can’t run forever if the result set keeps shifting. After the 3rd pass it logs a warning and stops. I also added a test that the job stops after 3 passes.

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -15,6 +15,10 @@
 
 from cds_rdm.inspire_harvester.transform.resource_types import ALL_DOCUMENT_TYPES
 
+INSPIRE_LITERATURE_API = "https://inspirehep.net/api/literature"
```
zzacharo: this should be a config, that we can easily adapt if e.g they change the API or we need to point to a test environment in our sandbox/dev

site/cds_rdm/inspire_harvester/reader.py
```diff
@@ -15,6 +15,10 @@
...
+# Cap re-harvest passes so a shifting result set cannot loop forever.
```
zzacharo: the shifting result is not understandable as a concept on its own....

site/cds_rdm/inspire_harvester/reader.py
kpsherva: the warning you added, is it resulting in partial success?

TahaKhan998: no, it is only a log warning. it does not set the run to partial success. partial success is when some entries have errors. after 3 passes we just stop, so the job can still be success even if we harvested fewer records than inspire reported.

TahaKhan998: now that i think about it, the run should be partial success if after 3 passes the missing records are not harvested so i added that now

PR #890, opened 2026-07-31
fix(search): show report numbers and EP badge in search results

### LINKED ISSUE #888: search: show report numbers in community search and EP numbers in results
On main search (`/search`), result items show the CDS report number, but the same records in a community search (`/communities/{slug}/records`) only show title and description. EP-approved records should also display their EP number (e.g. `CERN-EP-2026-001`) in search results on both main and community search, but that is missing today.

TahaKhan998:
Closes #888. Registers the CDS search result description override for community search so report numbers, accelerators, and experiments appear there as on main search, shows EP approval numbers (`apprn`) as a blue badge in the results label row on both, and limits each of those meta lists to two items with an ellipsis when there are more.

### DIFF
--- a/assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
+++ b/assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
@@ -7,7 +7,7 @@
 import _get from "lodash/get";
 import PropTypes from "prop-types";
 import React from "react";
-import { Item } from "semantic-ui-react";
+import { Item, Label } from "semantic-ui-react";
 
 export const CDSRecordsResultsListItemDescription = ({
   result,
@@ -28,13 +28,9 @@ export const CDSRecordsResultsListItemDescription = ({
     (acc) => acc.title.en
   );
 
-  const hasMetaData =
-    reportNumbers.length > 0 ||
-    cdsReferenceId ||
-    accelerators.length > 0 ||
-    experiments.length > 0;
-  const reportOrReference = reportNumbers.length > 0 || cdsReferenceId;
-  const referenceOrAccelerators = reportOrReference || accelerators.length > 0;
+  // CDS reference, accelerators, and/or experiments (row below report numbers).
+  const hasSideMeta =
+    Boolean(cdsReferenceId) || accelerators.length > 0 || experiments.length > 0;
 
   return (
     <>
@@ -42,40 +38,37 @@ export const CDSRecordsResultsListItemDescription = ({
         {descriptionStripped}
       </Item.Description>
 
-      {hasMetaData && (
-        <Item.Meta className="pt-20">
-          {reportNumbers.length > 0 && (
-            <span className="mr-5">
-              Report number:{" "}
-              {reportNumbers.map((reportNumber) => (
-                <span key={reportNumber} className="comma-separated">
-                  {reportNumber}
-                </span>
-              ))}
-            </span>
-          )}
-          {cdsReferenceId && (
-            <>
-              {reportNumbers.length > 0 && <span className="ml-5 mr-5">|</span>}
-              <span className={reportNumbers.length > 0 ? "ml-5 mr-5" : "mr-5"}>
-                {cdsReferenceId}
-              </span>
-            </>
-          )}
+      {reportNumbers.length > 0 && (
+        <Item.Meta className="pt-20 truncate-lines-1">
+          Report number: {reportNumbers.join(", ")}
+        </Item.Meta>
+      )}
+
+      {hasSideMeta && (
+        <Item.Meta
+          className={
+            reportNumbers.length > 0
+              ? "flex pt-5 search-result-meta-row"
+              : "flex pt-20 search-result-meta-row"
+          }
+        >
+          {cdsReferenceId && <span className="mr-5">{cdsReferenceId}</span>}
           {accelerators.length > 0 && (
             <>
-              {reportOrReference && <span className="ml-5 mr-5">|</span>}
-              <span className={reportOrReference ? "ml-5 mr-5" : "mr-5"}>
+              {cdsReferenceId && <span className="ml-5 mr-5">|</span>}
+              <div className="truncate-lines-1">
                 Accelerators: {accelerators.join(", ")}
-              </span>
+              </div>
             </>
           )}
           {experiments.length > 0 && (
             <>
-              {referenceOrAccelerators && <span className="ml-5 mr-5">|</span>}
-              <span className={referenceOrAccelerators ? "ml-5 mr-5" : "mr-5"}>
+              {(cdsReferenceId || accelerators.length > 0) && (
+                <span className="ml-5 mr-5">|</span>
+              )}
+              <div className="truncate-lines-1">
                 Experiments: {experiments.join(", ")}
-              </span>
+              </div>
             </>
           )}
         </Item.Meta>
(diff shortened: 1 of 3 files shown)


assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
```diff
@@ -6,13 +6,45 @@
...
+  }, [epNumber, result.id]);
+
+  return anchorRef;
+};
+
 export const CDSRecordsResultsListItemDescription = ({
   result,
   descriptionStripped,
 }) => {
+  const anchorRef = useEpNumberBadge(result);
```
zzacharo: why do we need this dynamic approach? i.e using anchors for the result?

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
```diff
@@ -6,13 +6,45 @@
 
 import _get from "lodash/get";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { useLayoutEffect, useRef } from "react";
 import { Item } from "semantic-ui-react";
 
+const getEpNumber = (result) =>
+  _get(result, "metadata.identifiers", []).find((id) => id.scheme === "apprn")
```
kpsherva: what about the EP report number in the main search (not community search)?

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
TahaKhan998: Upstream owns the top badge row, we only override the description under the title. We want the EP number as the last badge in that row, so this reaches into .labels-actions and appends it, similar to how the record page injects the EP badge in [detail.html](https://github.com/CERNDocumentServer/cds-rdm/blob/master/site/cds_rdm/templates/semantic-ui/cds_rdm/records/detail.html#L24-L34).

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
TahaKhan998: It’s covered for main search too. Same description override is registered for InvenioAppRdm.Search, so the EP badge shows there as well.

image

TahaKhan998:
image
Added limits to each of those meta lists to two items with an ellipsis when there are more.

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
```diff
@@ -6,13 +6,55 @@
...
 import _get from "lodash/get";
 import PropTypes from "prop-types";
-import React from "react";
+import React, { useLayoutEffect, useRef } from "react";
 import { Item } from "semantic-ui-react";
 
+/** Max items shown per meta list in search results (main + community). */
+const SEARCH_META_ITEMS_LIMIT = 2;
+
+const formatLimitedList = (items, limit = SEARCH_META_ITEMS_LIMIT) => {
```
zzacharo: hmmm this approach will be weird if you e.g have only report number but not the other fields shown and display an ellipsis while there is still space on the page.... @kpsherva thoughts?

sakshamarora1: Yeah we should go with ellipsis/truncation and not a hardcoded "config" for the number of report numbers, because report numbers can be long or short
Also, maybe it's better to move it to the end and not the beginning?
@TahaKhan998 look into how this can be via CSS (for [example](https://github.com/inveniosoftware/invenio-app-rdm/blob/700995a9ad2f5a7a02694d2f86d028eedc758bcd/invenio_app_rdm/theme/assets/semantic-ui/less/invenio_app_rdm/theme/globals/site.overrides#L433-L438))

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
sakshamarora1: Can we not use Overridables or parameterize?

TahaKhan998: Not really, we only override the description under the title and upstream owns the badge row. Theres no overridable or param there for the EP label so this was the smallest way to add it, same idea as detail.html.
We could override the full search result layout and render the badge in JSX but thats a bigger copy of upstream.

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
TahaKhan998: image
Done

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
zzacharo: Any way to upstream an overridable id for this?

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
zzacharo: What about if you add 10 accelerators or 10 experiments? Will the Report number label been shown up?

TahaKhan998: mobile:
image


desktop:
image

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
```diff
@@ -43,48 +42,70 @@ export const CDSRecordsResultsListItemDescription = ({
...
-          {experiments.length > 0 && (
-            <>
-              {referenceOrAccelerators && <span className="ml-5 mr-5">|</span>}
-              <span className={referenceOrAccelerators ? "ml-5 mr-5" : "mr-5"}>
-                Experiments: {experiments.join(", ")}
-              </span>
-            </>
+          {(cdsReferenceId || accelerators.length > 0 || experiments.length > 0) && (
+            <Item.Meta
+              className={
```
zzacharo: that seems quite complex.... wouldn't work to set a max-width and add the truncate-lines class to enable the ellpsis? @kpsherva ?

TahaKhan998: I tried the max-width + truncate-lines approach on the combined meta line. With many accelerators, the ellipsis cuts off before experiments, so that field gets eaten entirely. I went back to giving each part 50% max-width when both are present, so they stay on the same line and both remain visible when truncated.

kpsherva: I would prefer if we don't add custom CSS for something that was already addressed with truncate-lines in the past. We discussed to have one line for experiments + accelerators and separate line for report numbers. if you do for each truncate-lines-1, why doesn't it work?

TahaKhan998: I tried truncate-lines-1 on accelerators and experiments separately (same meta row). they no longer sit on one line and the | separator ends up alone on its own line.

So to keep them side by side we need the extra CSS. 
image

assets/js/components/search/overrides/CDSRecordsResultsListItemDescription.js
```diff
@@ -28,54 +28,46 @@ export const CDSRecordsResultsListItemDescription = ({
...
   );
 
-  const hasMetaData =
-    reportNumbers.length > 0 ||
-    cdsReferenceId ||
-    accelerators.length > 0 ||
-    experiments.length > 0;
-  const reportOrReference = reportNumbers.length > 0 || cdsReferenceId;
-  const referenceOrAccelerators = reportOrReference || accelerators.length > 0;
+  const hasSideMeta =
```
palkerecsenyi, line 32: maybe this could have a more descriptive name? I feel like `SideMeta` isn't super clear. or maybe just a quick comment to explain what it's doing

palkerecsenyi (approved):
looks great, thank you for this!! it will make the search results look a lot clearer. I just have one nitpick, feel free to ignore

kpsherva (approved):

PR #899, opened 2026-08-05
feat(harvester): use dedicated user with inspire-harvester role

### LINKED ISSUE #881: harvester: create a dedicated user for harvester task
- [x] we aim to have another user - to distinguish from system user, which we will use to run bulk scripts, but needs to have rights to create records and add to the research community
- [x] please coordinate the technical solution for this it with Zach

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/881
Moves the INSPIRE harvester off system_identity onto a dedicated service user with the inspire-harvester role, so harvested records have a clear owner. Harvester create/update/publish/search paths now use that identity.


site/cds_rdm/inspire_harvester/service_user.py
```diff
@@ -0,0 +1,52 @@
...
+    if hasattr(g, "cds_harvester_identity"):
+        return g.cds_harvester_identity
+
+    identity = get_identity(get_harvester_user())
+    if inspire_harvester_role not in identity.provides:
+        raise HarvesterUserError(
+            "Harvester user is missing the inspire-harvester role "
+            "(role id must equal name)."
+        )
+    g.cds_harvester_identity = identity
```
kpsherva: could you explain why did you choose to go with a global? it is an approach opposite to what we have normally in invenio

TahaKhan998: I cached the harvester user here because get harvester identity is called many times in a harvest run to create, publish, handle files, and search. If it is not cached then get harvester identity is called many times, which I thought could be slow. i removed it now to follow the invenio conventions

site/tests/conftest.py
```diff
@@ -621,6 +620,36 @@ def archiver(UserFixture, app, db):
...
 
+@pytest.fixture()
+def harvester_user(UserFixture, app, db):
+    """Dedicated INSPIRE harvester service user."""
+    ds = app.extensions["invenio-accounts"].datastore
+    user = UserFixture(
+        email=app.config["CDS_HARVESTER_USER_EMAIL"],
+        password="harvester",
+        preferences={
+            "visibility": "public",
```
kpsherva: please remember in real life the profile shouldn't be public

site/tests/conftest.py
```diff
@@ -621,6 +620,36 @@ def archiver(UserFixture, app, db):
...
+            "email_visibility": "restricted",
+            "notifications": {
```
kpsherva, line 647: same as above, in real life we don't want this user to receive any emails, do we?

site/cds_rdm/permissions.py
```diff
@@ -70,20 +71,33 @@ class CDSCommunitiesPermissionPolicy(CommunityPermissionPolicy):
 class CDSRDMRecordPermissionPolicy(RDMRecordPermissionPolicy):
     """Record permission policy."""
 
-    can_create = [AuthenticatedRegularUser(), SystemProcess()]
+    # Harvester role can manage records it does not own (legacy system-owned).
+    can_manage = RDMRecordPermissionPolicy.can_manage + [InspireHarvester()]
```
kpsherva: why should it be able to manage records?

site/cds_rdm/permissions.py
```diff
@@ -70,20 +71,33 @@ class CDSCommunitiesPermissionPolicy(CommunityPermissionPolicy):
...
     can_read = RDMRecordPermissionPolicy.can_read + [
         ArchiverRead(),
         CommitteeRefereeVersionGrant(),
+        InspireHarvester(),
+    ]
+    can_search = RDMRecordPermissionPolicy.can_search + [
+        ArchiverRead(),
+        InspireHarvester(),
+    ]
+    can_search_drafts = RDMRecordPermissionPolicy.can_search_drafts + [
```
kpsherva: why would it be able to search drafts?

site/cds_rdm/inspire_harvester/service_user.py
```diff
@@ -0,0 +1,48 @@
...
+
+def get_harvester_identity():
+    """Get an identity for harvester operations."""
+    identity = get_identity(get_harvester_user())
+    if inspire_harvester_role not in identity.provides:
+        raise HarvesterUserError(
+            "Harvester user is missing the inspire-harvester role "
+            "(role id must equal name)."
+        )
+    return identity
```
kpsherva: how about exploring an alternative implementation of these functions, in a form of a class with specific interface that allows us to retrieve id and identity?

site/cds_rdm/permissions.py
TahaKhan998: that was a misunderstanding, i thought the harvester was searching and updating drafts too. it doesn’t, so I’ll drop it.

site/cds_rdm/permissions.py
TahaKhan998: so we need to add the inspire harvester generator to the can_manage permission policy because previously many records were harvested by system and, in production, there are records owned by other users and not the inspire harvester user, so for it to be able to update those records with new info from inspire, create new versions of those records, or assign them to the right communities, we need can_manage.

site/cds_rdm/inspire_harvester/service_user.py
TahaKhan998: done i changed it to a class based implementation instead of separate helper functions so now one place handles getting the harvester user id and identity and i instantiate it once and reuse it in the flow so it stays cleaner and easier to maintain

site/cds_rdm/permissions.py
kpsherva: what about can_curate? isn't it enough? by assigning Harvester to manage you give it also permission to manage the access rights to the record. Doesn't it sound as too much for what the hervester does?

site/cds_rdm/inspire_harvester/service_user.py
```diff
@@ -0,0 +1,45 @@
...
+        email = current_app.config.get("CDS_HARVESTER_USER_EMAIL")
+        if not email:
+            raise RuntimeError("CDS_HARVESTER_USER_EMAIL is not configured.")
+
+        user = current_datastore.get_user_by_email(email)
+        if user is None:
+            raise RuntimeError(f"Harvester user '{email}' was not found.")
+        if not user.active:
+            raise RuntimeError(f"Harvester user '{email}' is inactive.")
+        return user
```
kpsherva: implementing the class like this will cause the DB query to be fired up every time you call the method - it is not very optimised approach, also considering that you call this method multiple times.

site/cds_rdm/inspire_harvester/service_user.py
```diff
@@ -0,0 +1,45 @@
...
+"""Harvester service user helpers."""
+
+from flask import current_app
+from invenio_access.utils import get_identity
+from invenio_accounts.proxies import current_datastore
+
+from cds_rdm.generators import inspire_harvester_role
+
+
+class HarvesterUserService:
```
kpsherva: normally this name is a bit misleading, since this is not really a service in invenio understanding

site/cds_rdm/inspire_harvester/service_user.py
```diff
@@ -0,0 +1,45 @@
...
+    """Service for resolving configured harvester user data."""
+
+    def user(self):
```
kpsherva: now the class starts to resemble something we already have existing which allows us to retrieve users - the users service, check https://github.com/inveniosoftware/invenio-users-resources/blob/master/invenio_users_resources/services/users/service.py

once you retrieve the user object, both id and the identity are reachable from this user object, so we don't need a wrapper class to help us retrieve it again. The only thing which we need is id parameter to be able to read the specific user

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -42,9 +42,13 @@ class InspireWriter(BaseWriter):
 
     def __init__(self):
         """Constructor."""
-        self.matcher = RecordMatcher()
-        self.drafts = DraftLifecycleManager()
-        self.file_sync = FileSynchronizer(draft_lifecycle=self.drafts)
+        self.harvester_user_service = HarvesterUserService()
+        self.identity = self.harvester_user_service.identity()
```
kpsherva: if you use user object mentioned in comments above, then this can be simplified


```suggestion
        self.harvester_user = HarvesterUser().get()
```

and later on you call (needs verification on your side)

```
self.harvester_user.identity
```

which is also more descriptive in the code, because we get to know of which user identity we are trying to call

site/cds_rdm/inspire_harvester/service_user.py
```diff
@@ -0,0 +1,45 @@
...
+        return user
+
+    def user_id(self):
+        """Get the configured harvester user id."""
+        return self.user().id
+
+    def identity(self):
+        """Get an identity for harvester operations."""
+        identity = get_identity(self.user())
+        if inspire_harvester_role not in identity.provides:
```
kpsherva: nit: this check is redundant - we should be informed when harvester tries to do something which it does not have permissions for by PermissionsError

palkerecsenyi (approved):
looks perfect!

zzacharo (approved):

PR #915, opened 2026-08-12
fix(harvester): error on withdrawn inspire records

### LINKED ISSUE #906: harverster feedback it2

TahaKhan998:
Part of https://github.com/CERNDocumentServer/cds-rdm/issues/906
if inspire marks a record as withdrawn we error and dont create/update it. if its already in cds we also put the cds id in the error so curators can open that record. if its not harvested yet we just say its withdrawn.

### DIFF
--- a/site/cds_rdm/inspire_harvester/transform/mappers/basic_metadata.py
+++ b/site/cds_rdm/inspire_harvester/transform/mappers/basic_metadata.py
@@ -338,3 +338,20 @@ class LanguagesMapper(MapperBase):
                     f"Failed mapping language. | details: lang={lang}, error={e}"
                 )
         return mapped_langs
+
+
+@dataclass(frozen=True)
+class WithdrawnMapper(MapperBase):
+    """Reject withdrawn INSPIRE records."""
+
+    id = "withdrawn"
+
+    def map_value(self, src_record, ctx, logger):
+        """Error if the INSPIRE withdrawn field is set."""
+        if not (src_record.get("metadata") or {}).get("withdrawn"):
+            return None
+        ctx.errors.append(
+            "Record is withdrawn on INSPIRE. "
+            f"| details: cds_id={ctx.cds_rdm_id}"
+        )
+        return None
--- a/site/cds_rdm/inspire_harvester/transform/config.py
+++ b/site/cds_rdm/inspire_harvester/transform/config.py
@@ -24,6 +24,7 @@ from cds_rdm.inspire_harvester.transform.mappers.basic_metadata import (
     ResourceTypeMapper,
     SubjectsMapper,
     TitleMapper,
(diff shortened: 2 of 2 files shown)


site/cds_rdm/inspire_harvester/transformer.py
```diff
@@ -36,6 +36,13 @@ def apply(self, stream_entry, **kwargs):
                 f"[INSPIRE#{control_number}] {error}" for error in unique_errors
             )
 
-        rdm_entry["_inspire_ctx"] = {"cds_id": cds_id, "versions": versions}
+        withdrawn = bool(
```
kpsherva: I would prefer to handle it like other fields, with a dedicated function and raise an error if the field exists - otherwise it is a bit "hidden" inside another code flow by not following the common convention

TahaKhan998: Done, added a dedicated WithdrawnMapper like the other fields.

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -98,14 +103,21 @@ def _route(self, stream_entry, inspire_id=None, record_pid=None, logger=None):
             stream_entry.errors.append(f"[inspire_id={inspire_id}] {msg}")
             return None
 
-        elif match_result.found:
+        if self._is_withdrawn(stream_entry):
```
kpsherva: why do we need the additional check? the mapper will throw an error anyways

TahaKhan998: well i thought that already harvested records with the withdrawn flag should be logged but i didnt realize that while being harvested, if the record has the withdrawn flag then the mapper would raise an error and we would just never reach the writer, thanks for catching this, i have dropped the writer error

site/cds_rdm/inspire_harvester/transform/mappers/basic_metadata.py
```diff
@@ -338,3 +338,23 @@ def map_value(self, src_record, ctx, logger):
...
+        """Error if the INSPIRE withdrawn field is set."""
+        if not (src_record.get("metadata") or {}).get("withdrawn"):
+            return None
+        if ctx.cds_rdm_id:
+            ctx.errors.append(
+                "Record is withdrawn on INSPIRE. "
+                f"| details: cds_id={ctx.cds_rdm_id}"
+            )
+        else:
+            ctx.errors.append("Record is withdrawn on INSPIRE.")
```
kpsherva: last tiny thing... why does it matter if it has cds id in this case or not? I would always raise. If the CDS PID is none, just display None :) 
in general I would not add conditional statements to modify log lines, since it does not handle bussiness logic in any way

Issue #908, opened 2026-08-07
move CERN-EP numbers to apprn schema


kpsherva:
We have migrated CERN-EP (check for other possible patterns of this number) as report number while it should be assigned as apprn
This has happened because there were records which were approved by EP but had no approval information in the metadata. The migration scripts will be modified accordingly.

- Review script in Gitlab: https://gitlab.cern.ch/cds-team/production_scripts/-/merge_requests/59

sakshamarora1:
Do you mean these 648 records?
https://repository.cern/search?q=metadata.identifiers.identifier%3A%22CERN-EP-%2A%22%20AND%20NOT%20metadata.identifiers.scheme%3A%22apprn%22&l=list&p=1&s=10&sort=bestmatch

Edit: Found a few more with PH and TH
https://repository.cern/search?q=%28metadata.identifiers.identifier%3A%22CERN-TH-%2A%22%20OR%20metadata.identifiers.identifier%3A%22CERN-EP-%2A%22%20OR%20metadata.identifiers.identifier%3A%22CERN-PH-%2A%22%29%20AND%20NOT%20metadata.identifiers.scheme%3A%22apprn%22&f=access_status%3Aopen&l=list&p=1&s=10&sort=bestmatch

Another question I have is that all of these records currently show the EP approval UI on the right side, I will need to check about that display as I'm not sure.

For example: https://repository.cern/records/yd5vp-31630

sakshamarora1:
Discussed IRL

1. Before the script, will need to update CDS-RDM to simplify the permissions for the validation in the component to only allow system user (admins cannot update via UI) plus we will not validate on the format of the report number. That should also fix the bug in #912

2. Check for how to make sure the right side panel of the EP request Panel would look for these records without a restricted record so that the listed records's report number moved from cdsrn to apprn doesn't break the UI.

sakshamarora1:
PR: needed before the script run: https://github.com/CERNDocumentServer/cds-rdm/pull/917
Migration Script: https://gitlab.cern.ch/cds-team/production_scripts/-/merge_requests/59

palkerecsenyi:
Looks good, I reviewed it on GitLab

zzacharo:
Update: discussed to IRL with @kpsherva to keep the validation on the cdsrn and move all EP like reportnumbers as apprn.

zzacharo:
ran on production

PR #917, opened 2026-08-12
fix(ep_approval): Fix error handling, add test for single record flow

### LINKED ISSUE #912: Bug: Manually added Approval report number vanishes when admin saves
https://github.com/user-attachments/assets/67ec80f7-8bea-4ec2-a723-f919374f4c33

Either we should allow admin (or just system to be able to run a script using service for adding it)
Or show an error

sakshamarora1:
Fixes: #912
Also makes the code ready for the migration script: https://gitlab.cern.ch/cds-team/production_scripts/-/merge_requests/59
Related: https://github.com/CERNDocumentServer/cds-rdm/issues/908


site/cds_rdm/components.py
```diff
@@ -177,7 +173,7 @@ def _validate_identifier_changes(self, identity, data, record):
         }
         if incoming_apprn != stored_apprn:
             error_msg = _(
-                "The 'apprn' identifier is system-managed and cannot be "
+                "The EP approval report number is system-managed and cannot be "
```
zzacharo: here we cannot hardcode the EP label as we would like to have this configurable. Maybe we added a label fro this in the config.

site/cds_rdm/components.py
```diff
@@ -201,59 +197,43 @@ def _validate_identifier_changes(self, identity, data, record):
...
+        reportnumber = committee_approval.get("reportnumber")
+        approved_internal = committee_approval.get("approved_internal_version")
+        source_internal = committee_approval.get("source_internal_version")
+        if not reportnumber or not source_internal:
+            return False
+        # Migrated case: both flags point at the same version, so only the current version carries apprn
+        if approved_internal == source_internal:
+            return record["id"] == source_internal
+        # Public copy: source points back to a different (internal) record, always sync
+        return True
```
zzacharo: isnt this always True?

site/cds_rdm/components.py
```diff
@@ -201,59 +197,43 @@ def _validate_identifier_changes(self, identity, data, record):
...
+    def _should_sync_apprn(self, record, committee_approval):
+        """Return True if apprn should be synced with parent committee_approval."""
+        reportnumber = committee_approval.get("reportnumber")
+        approved_internal = committee_approval.get("approved_internal_version")
+        source_internal = committee_approval.get("source_internal_version")
+        if not reportnumber or not source_internal:
+            return False
+        # Migrated case: both flags point at the same version, so only the current version carries apprn
+        if approved_internal == source_internal:
+            return record["id"] == source_internal
```
zzacharo: If I understand correctly, that wil regenerate the apprn only for the accepted version. The truth is that all the newer versions should have the apprn number

site/cds_rdm/requests/views.py
```diff
@@ -283,6 +283,7 @@ def publish_public_record(pid_value):
             pf["committee_approval"] = {
                 **ea,
                 "approved_public_version": new_record_id,
+                "source_public_version": src_id,
```
kpsherva, line 286: do we have documentation describing these keys?

site/cds_rdm/components.py
```diff
@@ -162,9 +158,9 @@ def _committee_approval_prefixes(self):
                 prefixes.add(prefix)
         return prefixes
 
-    def _validate_identifier_changes(self, identity, data, record):
-        """Raise ValidationError if the user is modifying protected identifiers."""
-        if self._is_privileged(identity):
+    def _validate_identifier_changes(self, identity, data, record, errors):
+        """Raise ValidationError if a non-system identity modifies apprn."""
+        if identity.id == system_user_id:
             return
```
sakshamarora1, line 164: Shouldn't this raise instead of return to keep inline with the docstring?

site/cds_rdm/requests/views.py
zzacharo: I will add some documentation in our internal docs

.invenio
```diff
@@ -13,7 +13,7 @@ description = CDS RDM InvenioRDM Instance
 author_name = CERN
 author_email = info@cds-rdm.com
 year = 2022
-python_version = 3.9
+python_version = 3.14
```
sakshamarora1, line 16: Is this leftover?

pyproject.toml
```diff
@@ -26,7 +26,8 @@ override-dependencies = [
     # therefore incompatible with Python <=3.9. See https://github.com/horejsek/python-fastjsonschema/issues/211
     # Their minimum Python version has now been set to 3.10 for >=2.22
     "fastjsonschema<2.22",
-    "invenio-assets<4.2.5"
+    "invenio-assets<4.2.5",
+    "flask-caching>=2.3.0,<2.5.0"
```
sakshamarora1: Is it related to this PR?

zzacharo: there is a separate commit pinning flask-cahcing, because tests were failing but it might be a leftover, I will check thanks!

.invenio
zzacharo: I am not sure if we are still using that, but I upgraded locally to 3.14 and upgraded the information there too.

site/cds_rdm/components.py
zzacharo: but we return if the user is system. We raise for non-system users as per docstring, do I miss something?

sakshamarora1: Yes, I mis-read, sorry!
But maybe we can have some kind of warning or log with the return, because the main bug might still happen if this component gets triggered by the system user either via a script or separately maybe due to a support ticket, and we don't return silently. WDYT?

zzacharo: wasn't a test added to test the fix of the initial bug?

sakshamarora1: I missed adding it before, we should add it.

site/cds_rdm/components.py
```diff
@@ -226,36 +222,44 @@ def _validate_identifier_changes(self, identity, data, record):
...
+
+        existing_identifiers = (data.get("metadata") or {}).get("identifiers", [])
+        not_apprn = [i for i in existing_identifiers if i.get("scheme") != "apprn"]
+
+        if self._should_sync_apprn(record, committee_approval):
+            reportnumbers = committee_approval.get("reportnumber") or []
+            apprn = [{"scheme": "apprn", "identifier": rn} for rn in reportnumbers]
+            new_identifiers = apprn + not_apprn
+        else:
+            new_identifiers = not_apprn
```
zzacharo, line 267: if the record does not have a `committee_approval` field then the identifiers are getting stripped out. That means, we need in the migration to populate the custom_field. This will be taken into account when running the https://gitlab.cern.ch/cds-team/production_scripts/-/merge_requests/59 and during migration.

zzacharo:
Screenshot 2026-09-03 at 17 00 51

site/cds_rdm/components.py
zzacharo: Shall we raise a validation error or warning when for some reason the `committee_approval` field is missing but apprn identifiers were passed instead of stripping them out? @kpsherva @palkerecsenyi @sakshamarora1

kpsherva: yes

sakshamarora1 (commented):
LGTM!
Screenshot 2026-09-04 at 16 15 38

zzacharo (approved):

PR #922, opened 2026-08-13
fix(harvester): do not create records for unmatched EP approvals

### LINKED ISSUE #906: harverster feedback it2

TahaKhan998:
part of https://github.com/CERNDocumentServer/cds-rdm/issues/906
If an INSPIRE record has an EP approval number and it does not match an existing CDS record, the harvester raises an error and does not create a new one. Same if it matches a restricted record, that means they forgot to publish the internal version after the approval. If it matches a public record we update it as usual.

### DIFF
--- a/site/cds_rdm/inspire_harvester/writer.py
+++ b/site/cds_rdm/inspire_harvester/writer.py
@@ -19,6 +19,7 @@ from marshmallow import ValidationError
 from cds_rdm.inspire_harvester.load.draft import DraftLifecycleManager
 from cds_rdm.inspire_harvester.load.files import FileSynchronizer
 from cds_rdm.inspire_harvester.load.matcher import RecordMatcher
+from cds_rdm.inspire_harvester.load.validator import RecordValidator
 from cds_rdm.inspire_harvester.logger import (
     Logger,
     format_validation_error,
@@ -45,6 +46,7 @@ class InspireWriter(BaseWriter):
         self.matcher = RecordMatcher()
         self.drafts = DraftLifecycleManager()
         self.file_sync = FileSynchronizer(draft_lifecycle=self.drafts)
+        self.record_validator = RecordValidator(self.matcher)
 
     def write(self, stream_entry, *args, **kwargs):
         """Create or update the record in CDS."""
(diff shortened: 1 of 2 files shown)


site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -99,11 +103,25 @@ def _route(self, stream_entry, inspire_id=None, record_pid=None, logger=None):
             return None
 
         elif match_result.found:
+            if apprn:
+                record = current_rdm_records_service.read(
+                    system_identity, match_result.record_pid
+                )
+                if record.to_dict().get("access", {}).get("record") == "restricted":
+                    raise WriterError(
+                        "EP approval number matched a restricted record. "
```
kpsherva:
```suggestion
                        "EP approval number matched a restricted record - record must be public to be updated by the harvester"
```

just to explain to the curators what should happen

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -99,11 +103,25 @@ def _route(self, stream_entry, inspire_id=None, record_pid=None, logger=None):
...
+                        f"| details: apprn={apprn}, cds_id={match_result.record_pid}"
+                    )
             logger.info(f"Matching record found: CDS#{match_result.record_pid}")
             self._update_record(stream_entry, record_pid=match_result.record_pid)
             return "update"
 
         else:
+            if apprn:
+                raise WriterError(
+                    "EP approval number did not match an existing record. "
```
kpsherva:
```suggestion
                    "EP approval number did not match an existing record - EP approval numbers can't be assigned outside CDS publishing workflow."
```

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -88,6 +88,10 @@ def _process_entry(self, stream_entry):
     def _route(self, stream_entry, inspire_id=None, record_pid=None, logger=None):
         """Route the entry to create or update based on existing record lookup."""
         match_result = self.matcher.match(stream_entry, inspire_id, logger)
+        apprn = self.matcher._retrieve_identifier(
+            stream_entry.entry.get("metadata", {}).get("identifiers", []),
+            "apprn",
+        )
```
kpsherva: these checks should not be implemented in the route step - if you also consider the other PR which you worked on the legacy recid matching - we will end up with a very unreadable method which has multiple responsibilities to handle - which is not the best practice.
these checks should have an additional validation class, organised similarly (composition) as the mappers - different validation workflow while updating the record and different while creating.

we are aiming to see, inside _update_record method something like
```
self.record_validator.validate(mode="update")
```
similar for the create mode

kpsherva (approved):

PR #926, opened 2026-08-14
fix(harvester): match experiment aliases from INSPIRE

### LINKED ISSUE #906: harverster feedback it2

TahaKhan998:
Part of https://github.com/CERNDocumentServer/cds-rdm/issues/906
INSPIRE harvests sometimes send experiment names that do not match our vocabulary ids (for example WA-095 vs WA95). After the existing exact-id lookup and hyphen/uppercase fallback, the harvester now searches props.aliases and uses the result when there is exactly one hit. Aliases were added on the experiment entries that were failing in harvest logs so those INSPIRE spellings map to the matching vocab ids.

### DIFF
--- a/site/cds_rdm/inspire_harvester/utils.py
+++ b/site/cds_rdm/inspire_harvester/utils.py
@@ -149,6 +149,12 @@ def get_vocabulary_exact(term, vocab_type, ctx, logger):
             if vocab_id:
                 return vocab_id
 
+        result = service.search(
+            system_identity, type=vocab_type, q=f'props.aliases.keyword:{term}'
+        )
+        if result.total == 1:
+            return list(result.hits)[0]["id"]
+
         logger.warning(
             f"Vocabulary term not found in '{vocab_type}'. | details: term={term}"
         )
--- a/app_data/vocabularies/experiments.yaml
+++ b/app_data/vocabularies/experiments.yaml
@@ -201,6 +201,8 @@
   description:
     en: 'A 4pi Solid Angle Detector for the SPS used as a Proton-Antiproton Collider
       at a Centre of Mass Energy of 540 GeV'
+  props:
+    aliases: UA-001
 - id: WA98
   title:
     en: WA98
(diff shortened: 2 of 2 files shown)


site/cds_rdm/inspire_harvester/utils.py
```diff
@@ -151,6 +149,12 @@ def get_vocabulary_exact(term, vocab_type, ctx, logger):
             if vocab_id:
                 return vocab_id
 
+        result = service.search(
+            system_identity, type=vocab_type, q=f'props.aliases:"{term}"'
```
kpsherva: did you check if search works also without specifying that we should search in the alias field (only by q:{term})?
what is the opensearch mapping assigned to this field once something is indexed? is it correct?

TahaKhan998: so basically i checked this locally with WA95, that one has alias WA-095. if i search just WA-095 with no field it finds WA95 but also other WA ones like WA98 and WA1 because of the hyphen. if i quote it i only get WA95 but that still searches the whole experiment row not only the alias, so if that string is in another experiment description we get more than one hit and we need one exact hit for it to work. like there is an experiment id ISOLDE and other experiments mention ISOLDE in their description (IS40, IS306) so searching "ISOLDE" with no field is not only the ISOLDE row. thats why the harvester searches props.aliases. Also opensearch can search it using props.aliases:"WA-095", these is no special mapping for it, its just a normal string like the other props.

kpsherva: all items indexed in opensearch have to have mapping assigned. What is the type assigned for the aliases in your local environment? string is not a type in opensearch. I am asking because based on the mapping, certain searches will be possible for example for "keyword" you will have to search for full alias, including the `-`, while other string types might match variations like to described. So in more rigid tokenisation, we will need to maintain all the possible aliases of an experiment, which is not ideal

TahaKhan998: switched to props.aliases.keyword:{term} for an exact match.

PR #931, opened 2026-08-18
fix(harvester): handle legacy CDS recid checks before create

### LINKED ISSUE #907: Harvester: avoiding duplicates between systems
when harvesting a record from inspire which has legacy recid (integer identifier) the harvester should:
        - check if the record is in repository (does it have legacy_recid minted in the pidstore?) if yes - great, proceed with harvesting, if not:
```
            - check if the record exists in the old system - if it redirects to repository.cern, it means it was migrated already but something went wrong with our pid table, raise an error to check this
            - if it does not exist in the old system - raise an error, means that there is something wrong with the identifiers we get from inspire
            - if it exists but does not redirect - don't harvest the record, add a warning that the record is still in the old system
```

think of how we could use other identifiers as well to detect and avoid duplication before the record is actually migrated

TahaKhan998:
Closes https://github.com/CERNDocumentServer/cds-rdm/issues/907
This PR changes how the INSPIRE harvester handles records with a legacy CDS recid and avoids duplicates between old CDS and CDS-RDM. If the incoming record has an integer legacy recid, we first check pidstore and update when it is found. If it is not found, we check old CDS directly. We raise an error when it redirects to repository.cern or returns 404. We skip with a warning when the record is still only in legacy. If there is no legacy recid in the incoming record, the existing matching flow stays the same. If INSPIRE sends more than one CDS recid (merged records keep both), we check each in pidstore and on old CDS. Two different pidstore hits are treated as multiple records match.

### DIFF
--- a/site/cds_rdm/inspire_harvester/load/matcher.py
+++ b/site/cds_rdm/inspire_harvester/load/matcher.py
@@ -10,9 +10,16 @@
 from dataclasses import dataclass, field
 from typing import List, Optional
 
+import requests
 from invenio_access.permissions import system_identity
 from invenio_rdm_records.proxies import current_rdm_records_service
 from invenio_search.engine import dsl
+from invenio_vocabularies.datastreams.errors import WriterError
+from sqlalchemy.orm.exc import NoResultFound
+
+from cds_rdm.inspire_harvester.utils import retrieve_identifiers
+from cds_rdm.legacy.resolver import get_pid_by_legacy_recid
+from cds_rdm.schemes import generate_cds_url, legacy_cds_pattern
 
 
 @dataclass
@@ -21,6 +28,7 @@ class MatchResult:
 
     ambiguous: bool = False
     found: bool = False
+    unmigrated: bool = False
     record_pid: Optional[str] = None
     matched_ids: List[str] = field(default_factory=list)
 
@@ -29,7 +37,18 @@ class MatchResult:
 class FilterCandidate:
     """Search filter tried as part of the record-matching priority chain."""
 
-    value: Optional[str]
+    values: object = None
+
+    def __post_init__(self):
+        """Cast ``values`` to a list so callers can pass ids as-is."""
+        raw = self.values
+        if isinstance(raw, str):
+            values = [raw]
+        elif raw:
+            values = [v for v in raw if v]
+        else:
+            values = self.values
+        object.__setattr__(self, "values", values)
 
     @property
     def query(self) -> List[dsl.Q]:
@@ -44,7 +63,7 @@ class ParentMatchFilter(FilterCandidate):
     @property
     def query(self):
         """Build the parent identifier query."""
-        return [dsl.Q("term", **{"parent.id": self.value})]
+        return [dsl.Q("terms", **{"parent.id": self.values})]
 
 
 @dataclass(frozen=True)
@@ -56,7 +75,7 @@ class CDSIdentifierMatchFilter(FilterCandidate):
         """Build the legacy CDS identifier query."""
         return [
             dsl.Q("term", **{"metadata.identifiers.scheme": "cds"}),
-            dsl.Q("term", **{"metadata.identifiers.identifier": self.value}),
+            dsl.Q("terms", **{"metadata.identifiers.identifier": self.values}),
         ]
 
 
@@ -67,7 +86,7 @@ class DOIMatchFilter(FilterCandidate):
     @property
     def query(self):
         """Build the DOI query."""
-        return [dsl.Q("term", **{"pids.doi.identifier.keyword": self.value})]
+        return [dsl.Q("terms", **{"pids.doi.identifier.keyword": self.values})]
 
 
 @dataclass(frozen=True)
@@ -80,8 +99,8 @@ class InspireIdentifierMatchFilter(FilterCandidate):
         return [
             dsl.Q("term", **{"metadata.related_identifiers.scheme": "inspire"}),
             dsl.Q(
-                "term",
-                **{"metadata.related_identifiers.identifier": self.value},
+                "terms",
+                **{"metadata.related_identifiers.identifier": self.values},
             ),
         ]
 
@@ -96,8 +115,8 @@ class ArxivIdentifierMatchFilter(FilterCandidate):
         return [
             dsl.Q("term", **{"metadata.related_identifiers.scheme": "arxiv"}),
             dsl.Q(
-                "term",
-                **{"metadata.related_identifiers.identifier": self.value},
+                "terms",
+                **{"metadata.related_identifiers.identifier": self.values},
             ),
         ]
 
@@ -112,8 +131,8 @@ class ReportNumberMatchFilter(FilterCandidate):
         return [
             dsl.Q("term", **{"metadata.identifiers.scheme": "cdsrn"}),
             dsl.Q(
-                "term",
-                **{"metadata.identifiers.identifier": self.value},
+                "terms",
+                **{"metadata.identifiers.identifier": self.values},
             ),
         ]
 
@@ -128,8 +147,8 @@ class RelatedReportNumberMatchFilter(FilterCandidate):
         return [
             dsl.Q("term", **{"metadata.related_identifiers.scheme": "cdsrn"}),
             dsl.Q(
-                "term",
-                **{"metadata.related_identifiers.identifier": self.value},
+                "terms",
+                **{"metadata.related_identifiers.identifier": self.values},
             ),
             dsl.Q(
                 "term",
@@ -148,8 +167,8 @@ class ApprovalReportNumberMatchFilter(FilterCandidate):
         return [
             dsl.Q("term", **{"metadata.identifiers.scheme": "apprn"}),
             dsl.Q(
-                "term",
-                **{"metadata.identifiers.identifier": self.value},
+                "terms",
+                **{"metadata.identifiers.identifier": self.values},
             ),
         ]
 
@@ -165,20 +184,26 @@ class RecordMatcher:
         """Search for existing records using a priority-ordered filter chain."""
         entry = stream_entry.entry
         ctx = entry["_inspire_ctx"]
+
+        migrated = self.match_migrated(entry, logger)
+        if migrated.found or migrated.ambiguous or migrated.unmigrated:
+            return migrated
+
         filter_candidates = self._build_filter_priority(
             entry, inspire_id, ctx["cds_id"]
         )
         result = None
         for candidate in filter_candidates:
-            if candidate.value:
-                combined_filter = dsl.Q("bool", filter=candidate.query)
-                logger.debug(f"Searching for existing records: {candidate.query}")
-                result = current_rdm_records_service.search(
-                    self.identity, extra_filter=combined_filter
-                )
-                if result.total >= 1:
-                    logger.debug(f"Found {result.total} matching records.")
-                    break
+            if not candidate.values:
+                continue
+            combined_filter = dsl.Q("bool", filter=candidate.query)
+            logger.debug(f"Searching for existing records: {candidate.query}")
+            result = current_rdm_records_service.search(
+                self.identity, extra_filter=combined_filter
+            )
+            if result.total >= 1:
+                logger.debug(f"Found {result.total} matching records.")
+                break
 
         if result is None or result.total == 0:
             return MatchResult(found=False)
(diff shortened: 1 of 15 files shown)


site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -103,9 +105,45 @@ def _route(self, stream_entry, inspire_id=None, record_pid=None, logger=None):
...
+            missing = []
+            for cds_id in cds_ids:
+                response = self._get_legacy_cds(cds_id)
+                location = response.headers.get("Location", "")
+                if "repository.cern" in location:
+                    redirected.append(cds_id)
+                elif response.status_code == 404:
+                    missing.append(cds_id)
+                else:
+                    still_on_legacy.append(cds_id)
```
kpsherva: why the errors are not raised straight away but are appended to the list? will the list ever have more than one item?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -103,9 +105,45 @@ def _route(self, stream_entry, inspire_id=None, record_pid=None, logger=None):
...
+                f"| details: recid={', '.join(missing)}"
+            )
+
+        self._create_record(stream_entry)
+        return "create"
+
+    def _get_legacy_cds(self, cds_id):
+        """Fetch old CDS without following redirects."""
+        url = generate_cds_url("cds", str(cds_id))
+        return requests.get(url, allow_redirects=False, timeout=15)
```
kpsherva: did you try with less costly HEAD request? is the response body needed for this check?

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -196,24 +186,31 @@ def _retrieve_identifier(self, identifiers, scheme) -> Optional[str]:
             None,
         )
 
+    def _legacy_cds_ids(self, entry) -> List[str]:
```
kpsherva: can we reuse retrieve_identifier function?

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -196,24 +186,31 @@ def _retrieve_identifier(self, identifiers, scheme) -> Optional[str]:
...
+        """Return integer legacy CDS identifiers from metadata identifiers."""
+        identifiers = entry.get("metadata", {}).get("identifiers", [])
+        return [
```
kpsherva: is there any advantage of using a list over following the generator pattern implemented above?

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -103,9 +105,45 @@ def _route(self, stream_entry, inspire_id=None, record_pid=None, logger=None):
...
+                logger.warning(
+                    "Record is still in the old CDS, skipping harvest. "
+                    f"| details: recid={', '.join(still_on_legacy)}"
+                )
+                return None
+            raise WriterError(
+                "INSPIRE CDS recid was not found in the old CDS. "
+                f"| details: recid={', '.join(missing)}"
+            )
+
```
kpsherva, line 124: doesn't this part deserve it's own method? the route method is responsible for dispatching specific behaviour depending on what the matcher found - in this case it is the matchers responsibility to tell us it matched a record which is still not migrated.
since the concept is to take a decision based on the matcher - you should aim to end up with something like 
```
elif matcher.unmigrated:
                logger.warning(
                    "Record is still in the old CDS, skipping harvest. "
                    f"| details: recid={', '.join(still_on_legacy)}"
                )
                return None
```
the other two WriterErrors should not be raised by router method but are matching errors

site/cds_rdm/inspire_harvester/writer.py
```diff
@@ -103,9 +105,45 @@ def _route(self, stream_entry, inspire_id=None, record_pid=None, logger=None):
...
+                    f"| details: recid={', '.join(redirected)}"
+                )
+            if still_on_legacy:
+                logger.warning(
+                    "Record is still in the old CDS, skipping harvest. "
+                    f"| details: recid={', '.join(still_on_legacy)}"
+                )
+                return None
+            raise WriterError(
+                "INSPIRE CDS recid was not found in the old CDS. "
```
kpsherva: this message is confusing because it mentions INSPIRE while we are matching against CDS recid in this case

site/cds_rdm/inspire_harvester/writer.py
kpsherva:
```suggestion
                if response.status_code == 404:
                    redirected.append(cds_id)
                elif response.status_code in ["301", "302"] and "repository.cern" in location:
                    missing.append(cds_id)
                elif response.status_code == 200:
                    still_on_legacy.append(cds_id)
                else:
                    raise ....
```
in order to make it a bit stricter on what behaviour is expected

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -161,6 +129,28 @@ def match(self, stream_entry, inspire_id, logger) -> MatchResult:
...
+                        system_identity, id_=parent_pid.pid_value
+                    )
+                    logger.debug(f"Found record by legacy recid {cds_id}: {record.id}")
+                    if record.id not in matched_ids:
+                        matched_ids.append(record.id)
+                except NoResultFound:
+                    logger.debug(f"No lrecid in pidstore for {cds_id}.")
+            if not matched_ids:
+                return MatchResult(found=False)
+            if len(matched_ids) > 1:
```
kpsherva: in which cases can this happen?

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -161,6 +129,28 @@ def match(self, stream_entry, inspire_id, logger) -> MatchResult:
...
+                return MatchResult(ambiguous=True, matched_ids=matched_ids)
+            return MatchResult(
+                found=True, record_pid=matched_ids[0], matched_ids=matched_ids
+            )
+
```
kpsherva, line 191: this part should be in separate method as you are matching for migrated record, it is different than a rest of the workflow so I would use it in more structured way, combined with what you have written in writer.py
i.e. here:
```
is_migrated = self.match_migrated(entry)
```

and 
```python
def match_migrated(self, entry):
     cds_ids = self._legacy_cds_ids(entry)
     if not cds_ids:
         return False

      for cds_id in cds_ids:
                try:
                    parent_pid = get_pid_by_legacy_recid(str(cds_id))
                    record = current_rdm_records_service.read_latest(
                        system_identity, id_=parent_pid.pid_value
                    )
                    logger.debug(f"Found record by legacy recid {cds_id}: {record.id}")
                    if record.id not in matched_ids:
                        matched_ids.append(record.id)
                except NoResultFound:
                    # here you check legacy if record migrated or not
                    # if not migrated you return something like MatchResult(unmigrated=True)
                    #
       if matched_ids > 1:
           # return ambiguous match if this is  possible to happen
       return MatchResult(
                found=True, record_pid=matched_ids[0], matched_ids=matched_ids
       )
```

site/cds_rdm/inspire_harvester/load/matcher.py
TahaKhan998: If the incoming INSPIRE record has two lrecids and they both point to different CDS RDM records, which is a rare case, but I thought it would be best to consider it too.

site/cds_rdm/inspire_harvester/load/validator.py
```diff
@@ -29,9 +29,12 @@ class EpApprovalCreateRule(ValidationRule):
 
     def check(self, stream_entry, *, record=None, record_pid=None, matcher=None):
         """Return an error if ``apprn`` is present on create."""
-        apprn = matcher._retrieve_identifier(
-            stream_entry.entry.get("metadata", {}).get("identifiers", []),
-            "apprn",
+        apprn = next(
+            matcher._retrieve_identifier(
```
kpsherva: I think _retrieve_identifier should be a utility function, not a class method, since it is being reused now by another class as well. Otherwise we are not isolating these classes but making them mutually dependent

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -189,24 +201,97 @@ def match(self, stream_entry, inspire_id, logger) -> MatchResult:
...
         doi = entry.get("pids", {}).get("doi", {}).get("identifier")
         identifiers = entry["metadata"].get("identifiers", [])
         related_identifiers = entry["metadata"].get("related_identifiers", [])
 
-        cds_id = self._retrieve_identifier(related_identifiers, "cds")
-        arxiv_id = self._retrieve_identifier(related_identifiers, "arxiv")
-        report_number = self._retrieve_identifier(identifiers, "cdsrn")
-        related_report_number = self._retrieve_identifier(related_identifiers, "cdsrn")
-        approval_report_number = self._retrieve_identifier(identifiers, "apprn")
+        cds_id = next(self._retrieve_identifier(related_identifiers, "cds"), None)
```
kpsherva: I am not sure why next is being called outside the retrieve_identifier function now? it would be more natural to keep it inside since you need to remember to call it each time when you use it.

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -189,24 +201,97 @@ def match(self, stream_entry, inspire_id, logger) -> MatchResult:
...
+
+        if matched_ids:
+            if len(matched_ids) > 1:
+                return MatchResult(ambiguous=True, matched_ids=matched_ids)
+            return MatchResult(
+                found=True, record_pid=matched_ids[0], matched_ids=matched_ids
+            )
+
+        if not unresolved:
+            return False
```
kpsherva:
```suggestion
        # after including next(.. inside the retrieve_identifiers and moving it to utility function)
        legacy_ids = [id for id in retrieve_identifiers(identifiers, "cds") if legacy_cds_pattern.match(str(id)]
        
        for legacy_id in legacy_ids:
            if not legacy_cds_pattern.match(str(cds_id)):
                continue
            try:
                parent_pid = get_pid_by_legacy_recid(str(cds_id))
                record = current_rdm_records_service.read_latest(
                    system_identity, id_=parent_pid.pid_value
                )
                logger.debug(f"Found record by legacy recid {cds_id}: {record.id}")
                if record.id not in matched_ids:
                    matched_ids.append(record.id)
            except NoResultFound:
                logger.debug(f"No lrecid in pidstore for {cds_id}.")
                unresolved.append(cds_id)

        if matched_ids:
            if len(matched_ids) > 1:
                return MatchResult(ambiguous=True, matched_ids=matched_ids)
            return MatchResult(
                found=True, record_pid=matched_ids[0], matched_ids=matched_ids
            )

        if not unresolved:
            # isnt it this instead to keep consistency on the returned type?
            return MatchResult(found=False)
```

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -189,31 +201,103 @@ def match(self, stream_entry, inspire_id, logger) -> MatchResult:
...
```
kpsherva: I think it was easily overlooked in my previous suggestion since I submitted the whole block of code:

```suggestion
            return MatchResult(found=False)
```

this is in order to keep the return type consistent.

TahaKhan998: yes just pushed this change

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -189,31 +201,103 @@ def match(self, stream_entry, inspire_id, logger) -> MatchResult:
...
+        if inspire_id:
+            candidates.append(InspireIdentifierMatchFilter(value=inspire_id))
+        for value in retrieve_identifiers(related_identifiers, "arxiv"):
+            candidates.append(ArxivIdentifierMatchFilter(value=value))
+        for value in retrieve_identifiers(identifiers, "cdsrn"):
+            candidates.append(ReportNumberMatchFilter(value=value))
+        for value in retrieve_identifiers(identifiers, "apprn"):
+            candidates.append(ApprovalReportNumberMatchFilter(value=value))
+        for value in retrieve_identifiers(related_identifiers, "cdsrn"):
+            candidates.append(RelatedReportNumberMatchFilter(value=value))
```
kpsherva: I would rather adapt the Filter classes to be able to consume lists and search for many results, because this approach is making the code a bit less readable

site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -189,31 +214,98 @@ def match(self, stream_entry, inspire_id, logger) -> MatchResult:
...
+            return MatchResult(unmigrated=True, matched_ids=still_on_legacy)
+        raise WriterError(
+            "CDS recid from INSPIRE was not found in the old CDS. "
+            f"| details: recid={', '.join(missing)}"
         )
 
+    def _get_legacy_cds(self, cds_id):
+        """Probe old CDS without following redirects."""
+        url = generate_cds_url("cds", str(cds_id))
+        return requests.head(url, allow_redirects=False, timeout=15)
```
kpsherva: last tiny thing: give the grandpa more time to answer, since we have issues with the system load due to bots :) I would set to 60

site/tests/inspire_harvester/test_harvester_job.py
```diff
@@ -78,9 +78,6 @@
         "title": "Fragmentation through Heavy and Light-flavor Measurements with the LHC ALICE Experiment",
         "publication_date": "2024",
         "languages": [{"id": "eng", "title": {"en": "English", "da": "Engelsk"}}],
-        "identifiers": [
```
kpsherva, line 81: I am not fully sure why this is removed from expected result - the results should still have CDS legacy recid inside the metadata, if this not how the records are created in the end then it might be incorrect - if record had legacy id and is ingested, it should be harvested with the legacy id in the metadata

TahaKhan998: we still map the cds recid into metadata.identifiers when it’s on the inspire record, this test is a create test though and after this pr if the fixture still has a numeric cds recid the matcher thinks it’s still on old cds and skips so nothing gets created and the test fails, that’s why we strip them from the mock inspire payload with drop_legacy_cds_ids and why they’re gone from expected_result, the input doesn’t have them anymore. migrated records still keep the legacy id, that’s what add_legacy_recid is for in the update tests.

zzacharo (approved):

PR #932, opened 2026-08-19
add(resource_types): letter and agenda

kpsherva:


app_data/vocabularies/resource_types.yaml
```diff
@@ -1133,6 +1152,25 @@
   tags:
     - depositable
     - linkable
+- id: publication-letter
```
zzacharo, line 1155: what's the usecase for that? I checked the Openaire definition for 0005 but this says "contribution to newspaper or weekly magazine"

kpsherva: we have in legacy: letter of intent (committees), letter (generic) for the DHOs (departmental content) and they are quite numerous cases, only for the former committees, it will be around 600 records. DHOs content, especially DG is mainly correspondence.
it is possible I assigned wrong openaire code (will check)

zzacharo: I agree with the first case but for correspondence isn't it wrong to map it to `publication-x` document type? Shouldn;t be distinguished?

kpsherva: we left the publication- prefix to be then displayed as text, because we didin't want to update all the records to change the resource type ids to 'text-...'

PR #950, opened 2026-08-31
fix(harvester): merge new abstracts and author identifiers on update

TahaKhan998:
Two things Fleur reported after editing records in INSPIRE didn't make it into CDS. The first one is the Italian abstract she added, which the mapper picks up fine but the update strategies didn't have an entry for additional descriptions, so it just got dropped, and now it's added as an append-only merge. The second one is the ORCID she added to the author, which got lost because each author gets one key for matching and it's the ORCID if there is one and the name otherwise, so the stored author was keyed by name and the incoming one by ORCID, they never matched and nothing was updated. Now every author lists all their keys, identifiers and name, and we try the identifiers first and fall back to the name, so an author who gains an ORCID upstream still matches the one we have and the ORCID gets merged in. The only exception is when both sides already have an ORCID and they're different, then it's two different people with the same name and we don't merge them.

### DIFF
--- a/site/cds_rdm/inspire_harvester/update/fields/creatibutors.py
+++ b/site/cds_rdm/inspire_harvester/update/fields/creatibutors.py
@@ -78,19 +78,36 @@ class CreatibutorsFieldUpdate(FieldUpdateBase):
 
         return out
 
-    def _key(self, creator: dict):
-        """Return a hashable key for matching a creator/contributor."""
+    def _keys(self, creator: dict):
+        """Return every key a creator/contributor may be matched on.
+
+        Identifiers come first, but the name key is always included so that an
+        author who gains an identifier upstream still matches the stored entry.
+        """
         p = creator.get("person_or_org") or {}
-        ids = p.get("identifiers") or []
-        for i in ids:
-            if i.get("scheme") and i.get("identifier"):
-                return ("id", i["scheme"], i["identifier"])
-        return (
-            "name",
-            (p.get("family_name") or "").lower(),
-            (p.get("given_name") or "").lower(),
-            (p.get("name") or "").lower(),
+        keys = [
+            ("id", i["scheme"], i["identifier"])
+            for i in p.get("identifiers") or []
+            if i.get("scheme") and i.get("identifier")
+        ]
+        keys.append(
+            (
+                "name",
+                (p.get("family_name") or "").lower(),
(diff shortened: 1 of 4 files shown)


site/cds_rdm/inspire_harvester/update/fields/creatibutors.py
```diff
@@ -78,19 +78,36 @@ def _union_affiliations(self, cur_list, inc_list):
...
+        keys = [
+            ("id", i["scheme"], i["identifier"])
+            for i in p.get("identifiers") or []
+            if i.get("scheme") and i.get("identifier")
+        ]
+        keys.append(
+            (
+                "name",
+                (p.get("family_name") or "").lower(),
+                (p.get("given_name") or "").lower(),
```
kpsherva, line 97: I am not sure I understand how this part will behave  - we can't be relying on matching only family name or only given name because it will lead to false positives, no?
ex. `Karolina ABC` and `Karolina EDF` are different names but if you use given name as matching key, they will be matched incorrectly.

TahaKhan998: So for your examples if INSPIRE sends Karolina ABC and Karolina EDF, theres no comma in the name so the mapper cant split it and the whole string just goes into family name. That gives you keys like ("name", "karolina abc", "", "") and ("name", "karolina edf", "", "") . Different keys so they dont match. If INSPIRE sends first name and last name separately instead you get ("name", "abc", "karolina", "abc, karolina") and ("name", "edf", "karolina", "edf, karolina"). Still different because the family name part is different even though given name is the same both times. So we are not matching on given name alone, the whole tuple has to match. 
What happens is it first tries to match on ORCID if there is one and only if that doesnt work it tries the name key. The case we're fixing is CDS has the author keyed by name only INSPIRE now has an ORCID and the ORCID lookup finds nothing but the name key matches so we merge and the ORCID gets added. But if both sides already have an ORCID and theyre different we dont fall back to the name even if it would match so two different people called Karolina dont get merged.

kpsherva (approved):

site/cds_rdm/inspire_harvester/update/fields/creatibutors.py
kpsherva: do we have tests covering different possible corner cases?

sakshamarora1 (approved):

PR #955, opened 2026-09-01
fix(harvester): do not map ISBN to imprint

### LINKED ISSUE #457: transformer: adjust imprint isbns custom field to non-thesis cases
- [x] Please check on how the ISBN mapping behaves for non thesis records. Is it assigned correctly in related identifiers with resource type book?
- [x] Are we mapping the identifiers and related identifiers in correct fields (compare with which schemas are configured in RDM_IDENTIFIERS_SCHEMES and RDM_RELATED_IDENTIFIER_SCHEMES)

More context under
https://github.com/CERNDocumentServer/cds-rdm/pull/444/files#r2075001463

TahaKhan998:
closes https://github.com/CERNDocumentServer/cds-rdm/issues/457
ImprintMapper was putting ISBNs into imprint.isbn, but after checking with Fleur, imprint doesn’t need an ISBN. So this PR stops mapping ISBN into imprint at all. All ISBNs still go to related_identifiers / Related Works, print, online, or no medium, that path was already fine. Added a test that imprint only keeps place and both print and online ISBNs land in related_identifiers.

I looked at IdentifiersMapper and RelatedIdentifiersMapper in identifiers.py, compared it to invenio.cfg, and basically only cds, aleph, cdsrn and apprn go to metadata.identifiers. Everything else like isbn, arxiv, urn, inspire, non CERN report numbers and extra dois go to metadata.related_identifiers, and the main doi goes to pids.doi, not those metadata fields. isbn used to go to identifiers wrongly but that’s already fixed, it only goes to related_identifiers. test_transformer.py covers this too.

### DIFF
--- a/site/tests/inspire_harvester/test_transformer.py
+++ b/site/tests/inspire_harvester/test_transformer.py
@@ -1210,27 +1210,6 @@ def test_transform_imprint_place(running_app):
     assert result["place"] == "Geneva"
 
 
-def test_transform_imprint_place_with_isbn(running_app):
-    """Test ImprintMapper with ISBN."""
-    src_metadata = {
-        "imprints": [{"place": "New York", "publisher": "Springer"}],
-        "isbns": [{"value": "978-3-16-148410-0", "medium": "online"}],
-        "control_number": 12345,
-    }
-    ctx = MetadataSerializationContext(
-        resource_type=ResourceType.OTHER, inspire_id="12345"
-    )
-    logger = Logger(inspire_id="12345")
-    mapper = ImprintMapper()
-    src_record = {"metadata": src_metadata, "created": "2023-01-01"}
-
-    result = mapper.map_value(src_record, ctx, logger)
-
-    assert "place" in result
-    assert result["place"] == "New York"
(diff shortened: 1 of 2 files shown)


site/tests/inspire_harvester/test_transformer.py
```diff
@@ -1210,25 +1210,39 @@ def test_transform_imprint_place(running_app):
...
 
-    result = mapper.map_value(src_record, ctx, logger)
+    imprint = ImprintMapper().map_value(src_record, ctx, logger)
+    related = RelatedIdentifiersMapper().map_value(src_record, ctx, logger)
 
-    assert "place" in result
-    assert result["place"] == "New York"
-    assert result["isbn"] == "978-3-16-148410-0"
+    assert imprint == {"place": "New York"}
+    assert "isbn" not in imprint
```
sakshamarora1: Is this check needed if we are already checking imprint against it's exact value just above this line?

TahaKhan998: yeah you're right, and the whole test is kinda pointless anyway since isbn is removed from ImprintMapper so a test isn't needed

PR #973, opened 2026-09-16
fix(harvester): keep shared DOI on the latest version

### LINKED ISSUE #943: harvester - validate harvests for multiple document types
- [x] check matching with migrated record without CDS DOI (shouldn't create version per resource type)
- [x] check creation of a new record without CDS DOI (should have one resource type version, the resource type should be assigned by priority
- [x] check updates of CDS DOIs record - does it create versions per resource type?
- [x] check creation of CDS DOIs record - shouldn't happen because we first need to have CDS record to even have a CDS DOI
- [x] check creation of subsequent versions - does it detect a new resource type correctly?
- [x] check updates of subsequent versions - does it update the correct version's metadata?
- [x] what happens when new files are added in all of the above cases?
- [x] think what else should we check

Provide examples of the behaviours above.

TahaKhan998:
closes https://github.com/CERNDocumentServer/cds-rdm/issues/943
When we split an INSPIRE record into versions and they share one journal DOI, both versions were getting that same pids.doi. The note is created first, so the article (the latest version) then fails with doi already exists.

What we do now: after the split, if an extra version has the same DOI as the latest one, we drop it from the extra version. Latest keeps it. A different DOI on an extra version is left alone.


site/cds_rdm/inspire_harvester/transform/splitter.py
```diff
@@ -15,6 +15,17 @@ def _is_arxiv(source: str) -> bool:
     return (source or "").lower() in _ARXIV_SOURCES
 
 
+def keep_shared_doi_on_latest(latest_entry, versions):
```
kpsherva: let's discuss this IRL with the team

kpsherva: more context - this is because of the external DOIs not being displayed on the latest version and if harvester adds a new version, the DOI will not be displayed on the latest
suggested approach: not to create versions for external DOIs, both in migration and harvester
ON HOLD

kpsherva: discussed:
when we are updating records with external DOIs, we should keep one version only (check how zenodo implements this - the feature is already there, it is a question of setting the permissions correctly)

TahaKhan998: updated. dropped the keep shared doi thing.
if theres no cds doi we just dont split. so one inspire record with two resource types and one journal doi stays one version, the priority type, and that one keeps the doi. no second version trying to take it.
didnt use the zenodo permission flag, that only blocks new version in the ui, it doesnt stop the harvester from building the sibling versions.

PR #975, opened 2026-09-16
feat(ep-approval): add read-only flag

### LINKED ISSUE #600: migration: evaluate communities with records with EP reportnumbers that they should not have the EP approval workflow enabled
Some old collections ended up with records with CERN EP reportnumbers although according to the EP committee they should not be enrolled for EP approval.

- Evaluate if we need to adjust the implementation to support this during migration and production
- Leave the config as-is and rely on the fact that if a cases arises there will be a request from their side which can be handled accordingly.

## Communities to disable future EP submissions

We need to apply these changes to the openshift environments

- ISOLDE
- LEP experiments

zzacharo:
read-only enrolled communities are displaying past approved records but disallow new submissions

closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/600

## Existing EP approved record
Screenshot 2026-09-16 at 15 34 39

## Non-requested EP approved record
Screenshot 2026-09-16 at 15 34 50


invenio.cfg
```diff
@@ -515,6 +515,17 @@ CDS_COMMITTEE_APPROVAL_COMMUNITIES = {
             "counter_digits": 3,  # zero-padding width, e.g. 3 → "001"
         },
     },
+    "1c46ceab-af68-4502-a436-cdf5a4cfbbe0": {
+        # isolde
+        "read_only": True,  # only allow read-only access to this community
```
kpsherva, line 524: you can add this as well to LEP experiments

site/cds_rdm/requests/committee_approval_state.py
```diff
@@ -222,9 +222,31 @@ def get_committee_approval_state(record_ui, record=None):
...
     open_request = _get_open_request(record_id, parent_record)
-    can_submit = _check_can_manage_community(community_id)
     approved_report_number = ea.get("reportnumber")
-    can_create_public = _check_can_create_public(can_submit, ea, record_id)
+
+    community_read_only = community_config.get("read_only", False)
+    if community_read_only:
+        if not approved_report_number:
+            return {
+                "can_submit": False,
```
kpsherva, line 231: could you explain why this dict need to be returned in this form? Out of curiosity, I want to understand how it works

invenio.cfg
zzacharo: we need to add this in our openshift environments too, lets document it in https://github.com/CERNDocumentServer/cds-migrator-kit/issues/600

site/cds_rdm/requests/committee_approval_state.py
zzacharo: basically this is a UI serialization on what is needed to render the [component](https://github.com/CERNDocumentServer/cds-rdm/blob/master/assets/js/components/record_details/CommitteeApproval.js#L18).

invenio.cfg
kpsherva: it would be good if you add it locally as well, for completeness and to avoid confusion

zzacharo: Can you check the updated list?

sakshamarora1 (approved):

PR #980, opened 2026-09-17
fix(harvester): conflict if thesis publication date changed

TahaKhan998:
Harvested records used to overwrite publication_date. For theses, a different incoming date is now a conflict and the update is skipped. Non-theses still overwrite.


site/cds_rdm/inspire_harvester/update/fields/metadata.py
```diff
@@ -11,11 +11,42 @@
...
+            return super().update(current, incoming, path, ctx)
+
+        if current_date and incoming_date and current_date != incoming_date:
+            return UpdateResult(
+                updated=current,
+                conflicts=[
+                    UpdateConflict(
+                        path=path,
+                        kind="date_mismatch",
+                        message="Incoming thesis publication_date differs",
```
kpsherva:
```suggestion
                        message="Incoming thesis publication_date differs, please check the date on INSPIRE",
```

kpsherva (approved):

PR #981, opened 2026-09-18
change(experiments): add alias for NA58

### LINKED ISSUE #492: migration: NA58 (COMPASS)
https://cds.cern.ch/collection/COMPASS?ln=en

palkerecsenyi:
Closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/492

---


app_data/vocabularies/experiments.yaml
```diff
@@ -342,6 +342,7 @@
     en: 'Common Muon and Proton Apparatus for Structure and Spectroscopy'
   props:
     link: http://wwwcompass.cern.ch/
+    aliases: COMPASS NA58
```
kpsherva, line 345: shouldn't be only COMPASS? NA58 is already in the id

palkerecsenyi: Many records on legacy CDS have `693__e:"COMPASS NA58"` and some of the collections have that in their configured query (see "COMPASS Conference Proceedings" for example). We put the alias here just so that it can match the legacy experiment to an RDM experiment (as well as for matching experiments from INSPIRE), it is not e.g. shown in the UI anywhere.

zzacharo: Having only COMPASS should also work as it searches in both id and aliases no?

palkerecsenyi: Right now the `aliases` field only actually supports one alias (both for INSPIRE and the migrator kit). So `COMPASS NA58` is only one value, it is not e.g. a space-separated list. [When matching](https://github.com/CERNDocumentServer/cds-migrator-kit/blob/a40efd739a62e1265c620c6ae97c33f962443234/cds_migrator_kit/rdm/records/transform/mappers/vocabulary.py#L59), it will match either `NA58` as the ID, then `NA58` as the title, and then `COMPASS NA58` if an ID or title match can't be found.

In legacy, most of the records have a `693__e` of `COMPASS NA58` but a small number of them have `COMPASS` and some have `NA58`. So I need to update my PR to support just `COMPASS` too, which might need some changes to the migrator kit / INSPIRE harvester.

PR #986, opened 2026-09-21
fix(harvester): match parent DOI as well as version DOI

### LINKED ISSUE #347: INSPIRE to CDS-RDM harvester: sync document types already in CDS
A few collaborations are already using the new CDS for documents other than Thesis.
After the completion of the sync of Thesis with INSPIRE, we want to sync these existing documents.

We want to:
- Identify what are the documents already submitted to the new CDS (FCC physics, others?) and potentially submitted to other external repositories.
- Update such documents with modifications coming from INSPIRE. This is non-trivial, as INSPIRE combines multiple document types (e.g. Preprint and Articles) in the same record. We need to define the criteria (what field values) that identify only the collaborations that are using the new CDS.
- Insert ew records only for the identified criteria. We should avoid creating new records for the same collaboration in the current and new CDS (duplications).
- ~~Modify the harvester in the current CDS to skip the sync for already migrated records.~~ DONE
- Check if there are already duplicated records, coming from INSPIRE, in the current CDS that should be in the new CDS.

TahaKhan998:
closes https://github.com/CERNDocumentServer/cds-rdm/issues/347
The matcher only searches pids.doi (the version DOI). For some hand-submitted records INSPIRE has the parent DOI instead, so we miss the existing CDS record and create a duplicate. DOIMatchFilter now also searches parent.pids.doi, so either DOI is enough to match.


site/cds_rdm/inspire_harvester/load/matcher.py
```diff
@@ -81,12 +81,24 @@ def query(self):
...
-        return [dsl.Q("terms", **{"pids.doi.identifier.keyword": self.values})]
+        """Build the DOI query for version and parent pids."""
+        return [
+            dsl.Q(
+                "bool",
+                should=[
+                    dsl.Q("terms", **{"pids.doi.identifier.keyword": self.values}),
+                    dsl.Q(
+                        "terms",
+                        **{"parent.pids.doi.identifier.keyword": self.values},
```
kpsherva, line 96: could you add a test to visualise what happens if we match a parent?

site/tests/inspire_harvester/test_matcher.py
```diff
@@ -9,14 +9,62 @@
...
 from invenio_vocabularies.datastreams.errors import WriterError
 from sqlalchemy.orm.exc import NoResultFound
 
-from cds_rdm.inspire_harvester.load.matcher import RecordMatcher
+from cds_rdm.inspire_harvester.load.matcher import DOIMatchFilter, RecordMatcher
 
 from .utils import legacy_entry
 
 
+def test_doi_match_filter_searches_version_and_parent_doi():
```
kpsherva: what will happen in terms of create/update? will it work properly? which version will be updated if the parent is found?

TahaKhan998: well if no match is found then it creates a new record, if the parent doi matches then the search only returns the latest version by default and that is what gets updated. if we want a specific version to be updated then we could find the specific version using the resource type of the incoming inspire record maybe?

PR #987, opened 2026-09-21
fix(harvester): allow sandbox create of missing CDS DOI records

### LINKED ISSUE #978: harvester: allow create in sandbox when the CDS record is not in the local DB
In sandbox the harvester tries to create records that already exist on prod. those records have a CDS DOI (10.17181) and/or a CDSRDM id. the matcher finds nothing in sandbox because the record is not there, so it goes to create. create is then blocked with:

"Create is blocked: trying to create record with cds doi - record should be updated instead."

this only happens because sandbox does not have the same records as prod. 
add a CDS config (sandbox only). when the matcher finds no local hit, allow create even if the incoming record has a datacite DOI or a CDSRDM id. use the same CDS pid as prod.

TahaKhan998:
closes #978

in sandbox the matcher finds nothing for records that already exist on prod (they have a CDS DOI). create then gets blocked.
added CDS_HARVESTER_ALLOW_CDS_DOI_CREATE (off by default). when on, allow create and keep the same CDS DOI. sandbox config sets this to True.


site/cds_rdm/inspire_harvester/load/validator.py
```diff
@@ -52,6 +52,8 @@ class CdsDoiCreateRule(ValidationRule):
 
     def check(self, stream_entry, *, record=None, record_pid=None, matcher=None):
         """Return an error if the entry DOI uses the CDS DataCite prefix."""
+        if current_app.config["CDS_HARVESTER_ALLOW_CDS_DOI_CREATE"]:
```
kpsherva: shouldn't this also apply for cds-rdm identifier as we discussed, not only the DOI?

kpsherva: if we have CDS-RDM only, no doi, is the new record created with the same PID assigned to the record and the parent(this might come later if we don't receive parent pids from inspire yet)?

TahaKhan998: the config covers cdsrdm now too. if inspire sends a cdsrdm id and the config is true, we reuse it as the parent recid on create. the version still gets a new id because parent and version cant be the same recid.
if its cdsrdm only with no doi, create already went through before. with the config true we also keep that same parent id instead of minting a new one.

renamed the config to CDS_HARVESTER_ALLOW_MISSING_CDS_CREATE so it is not doi only.

site/cds_rdm/inspire_harvester/load/draft.py
```diff
@@ -27,9 +29,40 @@ def __init__(self, identity):
...
-        """Create a new draft from entry data."""
-        return current_rdm_records_service.create(self.identity, data=entry)
+    def create(self, entry, parent_pid=None):
+        """Create a new draft from entry data.
+
+        :param parent_pid: Optional CDSRDM/parent recid to reuse instead of
+            minting a new parent id (sandbox create of a prod record).
+        """
+        draft = current_rdm_records_service.create(self.identity, data=entry)
+        if parent_pid:
```
kpsherva, line 39: are you 100% sure that what we receive from inspire is the parent pid? or this is the record pid?
we can't do this create operation in production though, if there is any discrepancy between CDS and INSPIRE (wrong PID for example), this will start creating wrong records in prod, it should be only in non-prod environments.
in addition, please use the unit of work to encapsulate these operations as one atomic one

